Issue #1, opened 2021-03-10
Write Unit tests and API tests


twobeeb:
Current code requires Unit tests as well as Integration tests
See both client and controllers examples in current master for API and Unit tests examples 
Any other suggestion for easier test definition/coverage welcome

PR #32, opened 2021-03-30
Create Unit Tests

Debatent:
#1


twobeeb:
Use ``Mockito.lenient()`` in @BeforeEach if the call you are mocking is not necessary for every test case
https://github.com/mockito/mockito/issues/1540#issuecomment-557850819

PR #56, opened 2021-05-28
Feat/synchronize namespace

seb63110:
Iterate the accessControlEntries of ns4kfk for a namespace and create topics and connects into ns4kfk for existing topics and connects on the cluster


twobeeb (approved):
:thumbsup:

PR #81, opened 2021-06-28
Increased init-timeout to consume all messages to 60s

seb63110:


twobeeb:
At the startup of ns4kafka, a thread is created to consume compacted topic (1thread per topic)
Process was timeboxed to 10s max.
Once init timeout is reached, exception is thrown but since it's inside a thread, it's not caught and kafkactl remains stuck in infinite waiting loop.
This fix doesn't remove the uncaught exception but : 
- better informs users of what's happening during startup, 
- as well as increase the timeout to process all messages to 60s by default

Issue #10, opened 2021-03-12
Log events that led to modification of resources


twobeeb:
Track:
- Who made the change
- object before
- object after
- metadata
- other ?

PR #82, opened 2021-06-30
Feat/log user action

Debatent:


Debatent:
#10

twobeeb (changes requested):
1. I'm ok with Lombok initiative 
2. Logging before each controller method is a NO : 
    - doesn't bring anything useful (especially in INFO level)
    - already provided by micronaut if anyone needs it https://docs.micronaut.io/latest/guide/#accessLogger

Issue #83, opened 2021-07-04
Provide to kafakctl the created/changed/unchanged status


twobeeb:
Currently ns4kafka API does not provide a mechanism to understand if the object was changed or unchanged as a result of the apply.

While documenting the README.md I described the output of kafkactl following 2 consectutive apply as such :  
````console
user@local:/home/user$ kafkactl apply -f topic.yml
Success Topic/test.topic1 (created)
# deploy twice
user@local:/home/user$ kafkactl apply -f topic.yml
Success Topic/test.topic1 (unchanged)
````

In order to replicate this exact behavior, we need ns4kafka API to provide in its `Success` response a new header `X-Ns4kafka-Result` with 3 possibles values : 
- `created` when the resource did not exist beforehand
- `changed` when the resource Spec or Label got updated
- `unchanged` otherwise

Some amount of work must also be done to align the console output to better match what was documented inthe README.md

PR #84, opened 2021-07-07
Improved Error Handling and new Status header on apply

Debatent:
#83


twobeeb (approved):
Minors changes required on CLI side to better handle errors but overall OK

Issue #102, opened 2021-08-27
Add the possibility to Restart Connector from Ns4kafka


Debatent:
As Ns4kafka manages the clusters Kafka, it seems relevant to add a Administration command to restart the Connector.
From the CLI the command should look like this:

kafkactl restart-connect connector-name

PR #103, opened 2021-08-27
Feat/restart connect

Debatent:
#102
````
Usage: kafkactl connectors <action> <connectors>...
Interact with connectors (Pause/Resume/Restart)
      <action>          (pause | resume | restart)
      <connectors>...   Connector names separated by space (use `ALL` for all connectors)

````


twobeeb (changes requested):
2 things to change : 
- kafka connect restart doesn't use the url you defined (restart must happen at task level : https://docs.confluent.io/platform/current/connect/references/restapi.html#post--connectors-(string-name)-tasks-(int-taskid)-restart)
- don't use custom connector (sound connector)

PR #194, opened 2022-07-06
Perform retry when ns4kafka timeout on schema registry/kconnect

loicgreffier:


loicgreffier:
Hi @twobeeb, 

The issue we are facing right now seems more related to network than Ns4Kafka itself:
- From trace logs, the first HTTP call to Ns4Kafka itself (on SCHEMA_REGISTY_PROXY or CONNECT_PROXY) is always sent.
- The redirection made by the internal HTTP filter has been always sent.
- Timeout happens after 10 seconds (default value). When trying to increase this timeout (e.g. to 5 minutes), something is responding HTTP 500:
  - There are no trace logs in Ns4kafka showing something wrong happened. We can only see a response 500 is received.
  - Requests never arrive to the recipient (neither schema registry nor Kafka Connect)
- It happens randomly (e.g. submitting 300 schemas can work perfectly as it can give 2 or 3 requests timeout) and I never managed to reproduce the issue on a local workstation

I already considered reactive as a solution, but correct me if I'm wrong: this would imply a whole rewriting of the schema/connector controllers to chain reactive calls (from controllers, to services, to repositories) ? That's why we go to this quick fix retry mechanism, at least, for now

PR #225, opened 2022-12-22
Allowed use of wildcard as ACLs grantedTo specification to manage public topics

yraffin:


api/src/main/java/com/michelin/ns4kafka/controllers/AccessControlListController.java
```diff
@@ -59,7 +61,7 @@ public List<AccessControlEntry> list(String namespace, Optional<AclLimit> limit)
                 return accessControlEntryService.findAllGrantedToNamespace(ns)
                         .stream()
                         // granted to me
-                        .filter(accessControlEntry -> accessControlEntry.getSpec().getGrantedTo().equals(namespace))
+                        //.filter(accessControlEntry -> accessControlEntry.getSpec().getGrantedTo().equals(namespace))
```
loicgreffier, line 64: You can remove the comment + the commented line

PR #231, opened 2023-01-10
Adding /acls endpoint to get all the ACLs

AlexisSouquiere:


src/main/java/com/michelin/ns4kafka/controllers/acl/AccessControlListNonNamespacedController.java
```diff
@@ -0,0 +1,34 @@
...
+     */
+    @Inject
+    AccessControlEntryService accessControlEntryService;
+
+    /**
+     * Get all the ACLs of all namespaces
+     * @return A list of ACLs
+     */
+    @Get
+    public List<AccessControlEntry> listAll() {
```
loicgreffier, line 31: Add unit tests for this new controller (see: [ResourceQuotaNonNamespacedControllerTest.java](https://github.com/michelin/ns4kafka/blob/master/src/test/java/com/michelin/ns4kafka/controllers/ResourceQuotaNonNamespacedControllerTest.java) for example)

src/main/java/com/michelin/ns4kafka/services/AccessControlEntryService.java
```diff
@@ -271,6 +271,14 @@ public List<AccessControlEntry> findAllForCluster(String cluster) {
                 .collect(Collectors.toList());
     }
 
+    /**
+     * Find all the ACLs on all clusters
+     * @return A list of ACLs
+     */
+    public List<AccessControlEntry> findAll() {
```
loicgreffier, line 278: Add unit tests for this new method (see: [TopicServiceTest](https://github.com/michelin/ns4kafka/blob/master/src/test/java/com/michelin/ns4kafka/services/TopicServiceTest.java#L916) for example)

PR #245, opened 2023-01-27
Added AES256 vault capability for Kafka Connect clusters

yraffin:


src/main/java/com/michelin/ns4kafka/controllers/ConnectClusterController.java
```diff
@@ -139,4 +147,61 @@ public HttpResponse<Void> delete(String namespace, String connectCluster, @Query
         connectClusterService.delete(connectClusterToDelete);
         return HttpResponse.noContent();
     }
+
+    /**
+     * Gets the list of available Kafka Connect vaults for a namespace
+     *
+     * @return The list of the available Kafka Connect vaults.
+     */
+    @Get("/vaults")
```
loicgreffier, line 156: `@Get("/_/vaults")` is required (see `imports` endpoint) to match role bindings criteria

PR #260, opened 2023-02-24
Added producer/consumer byte rate client quota customization

AlexisSouquiere:
- Adding client quota customization
```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: myNamespace
  cluster: cluster1
spec:
  kafkaUser: kafkaServiceAccount
  kafkaUserQuota:
    producer_byte_rate: 204800
    consumer_byte_rate: 409600
```

- Fixing update user quota filtering. Update has to be performed if namespace user quota definition is different from the broker user quota
- Adding integration tests for both custom quota and quota updates

**Note**: at the beginning I was thinking of specifying the user quota like this
```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: myNamespace
  cluster: cluster1
spec:
  kafkaUser: 
    name: kafkaServiceAccount
    producer_byte_rate: 204800
    consumer_byte_rate: 409600
```
But this format would have broken the namespace unmarshalling with existing namespace definition.
I was also thinking of creating a apiVersion 2 for namespace definition, Namespace v2 model, etc. but kept a simplier approach by adding a new configuration key. 
Tell me what you prefer... ! 

When it will be merged I'll created a PR oin kafkactl to update to Readme accordingly

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/services/ResourceQuotaService.java
+++ b/src/main/java/com/michelin/ns4kafka/services/ResourceQuotaService.java
@@ -6,6 +6,7 @@ import com.michelin.ns4kafka.models.Topic;
 import com.michelin.ns4kafka.models.quota.ResourceQuota;
 import com.michelin.ns4kafka.models.quota.ResourceQuotaResponse;
 import com.michelin.ns4kafka.repositories.ResourceQuotaRepository;
+import com.michelin.ns4kafka.services.executors.UserAsyncExecutor;
 import com.michelin.ns4kafka.utils.BytesUtils;
 import io.micronaut.core.util.StringUtils;
 import jakarta.inject.Inject;
@@ -27,6 +28,7 @@ public class ResourceQuotaService {
     private static final String QUOTA_PARSE_ERROR = "Number expected for %s (%s given)";
 
     private static final String QUOTA_RESPONSE_FORMAT = "%s/%s";
+    private static final String USER_QUOTA_RESPONSE_FORMAT = "%sB/s";
 
     private static final String NO_QUOTA_RESPONSE_FORMAT = "%s";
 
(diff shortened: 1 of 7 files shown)


AlexisSouquiere:
Hi @loicgreffier

You're right these properties should be in the ResourceQuota resource. My solution was straightforward (no need to do an extra call to the ResourceQuotaRepository because we had everything we need in the namespace definition) but functionally your proposal is better.

I'm finishing some tests and will update the PR

AlexisSouquiere:
@loicgreffier can you check again the PR please ?

src/main/java/com/michelin/ns4kafka/services/ResourceQuotaService.java
```diff
@@ -127,6 +127,24 @@ public List<String> validateNewResourceQuota(Namespace namespace, ResourceQuota
             }
         }
 
+        String producerByteRate = resourceQuota.getSpec().get(USER_PRODUCER_BYTE_RATE.getKey());
```
loicgreffier, line 130: Could you add a unit test in `ResourceQuotaServiceTest` for these new conditions ?

loicgreffier:
Hi @AlexisSouquiere, in addition, could these new quotas be returned by `list` ([link](https://github.com/michelin/ns4kafka/blob/master/src/main/java/com/michelin/ns4kafka/controllers/quota/ResourceQuotaController.java#L37)) and `get` ([link](https://github.com/michelin/ns4kafka/blob/master/src/main/java/com/michelin/ns4kafka/controllers/quota/ResourceQuotaController.java#L50)) in quota controller, so we can then display them in [kafkactl](https://github.com/michelin/kafkactl) ?

src/main/java/com/michelin/ns4kafka/services/ResourceQuotaService.java
AlexisSouquiere: Done

AlexisSouquiere:
I did the changes to return these 2 new quotas in the list/get endpoints. If it's ok I'll create the PR for the same on kafkactl (https://github.com/AlexisSouquiere/kafkactl/commit/b83c764c6330ee142f132fc56682b9e11e329009)

loicgreffier:
@AlexisSouquiere, OK for merge, you can open a PR on Kafkactl

PR #344, opened 2023-11-28
Support unchanged status on schemas

AlexisSouquiere:


src/main/java/com/michelin/ns4kafka/controllers/SchemaController.java
```diff
@@ -100,52 +101,63 @@ public Mono<HttpResponse<Schema>> apply(String namespace, @Valid @Body Schema sc
...
+                                            validationErrors,
+                                            schema.getKind(),
+                                            schema.getMetadata().getName()));
+                                }
+
+                                schema.getMetadata().setCreationTimestamp(Date.from(Instant.now()));
+                                schema.getMetadata().setCluster(ns.getMetadata().getCluster());
+                                schema.getMetadata().setNamespace(ns.getMetadata().getName());
+
+                                if (dryrun) {
```
loicgreffier, line 133: I think this:

```java
if (dryrun) {
  return Mono.just(formatHttpResponse(schema,
          latestSubjectOptional.isPresent()
                  ? ApplyStatus.changed : ApplyStatus.created));
}
  
return schemaService
      .register(ns, schema)
      .map(id -> {
          ApplyStatus status;
  
          if (latestSubjectOptional.isEmpty()) {
              status = ApplyStatus.created;
              sendEventLog(schema.getKind(), schema.getMetadata(), status, null,
                  schema.getSpec());
          } else if (id > latestSubjectOptional.get().getSpec().getId()) {
              status = ApplyStatus.changed;
              sendEventLog(schema.getKind(), schema.getMetadata(), status,
                  latestSubjectOptional.get().getSpec(),
                  schema.getSpec());
          } else {
              status = ApplyStatus.unchanged;
          }
  
          return formatHttpResponse(schema, status);
      });
});
```

could be now simplified to this:

```java
ApplyStatus status = latestSubjectOptional.isPresent() ? ApplyStatus.changed : ApplyStatus.created;
if (dryrun) {
    return Mono.just(formatHttpResponse(schema, status);
}

return schemaService
    .register(ns, schema)
    .map(id -> formatHttpResponse(schema, status));
```

Since we now use the canonical string to compute the status, we do not need the schema registry response for the final HTTP response, what do you think ? I'm gonna try this out

src/main/java/com/michelin/ns4kafka/controllers/SchemaController.java
```diff
@@ -100,52 +101,63 @@ public Mono<HttpResponse<Schema>> apply(String namespace, @Valid @Body Schema sc
...
-                if (!validationErrors.isEmpty()) {
-                    return Mono.error(new ResourceValidationException(validationErrors, schema.getKind(),
-                        schema.getMetadata().getName()));
-                }
+                .getLatestSubject(ns, schema.getMetadata().getName())
+                .map(Optional::of)
+                .defaultIfEmpty(Optional.empty())
+                .flatMap(latestSubjectOptional -> {
+                    if (latestSubjectOptional.isPresent()) {
+                        var newSchema = new AvroSchema(schema.getSpec().getSchema());
```
twobeeb, line 109: There's a constructor for `AvroSchema` that takes schema references, here you take a risk to compare without references later in `canonicalString()`. 
It was done automatically on the registry side before

Issue #146, opened 2022-01-05
Optimize JWT content


twobeeb:
Currently the JWT generated during authentication contains the list of the user's groups.
This list can grow quite big depending on the authentication mecanism (gitlab, LDAP, ...)
````json
{
  "sub": "julien.chanaud",
  "nbf": 1641311354,
  "roles": [
    "isAdmin()"
  ],
  "iss": "ns4kafka",
  "groups": [
    "project/group1",
    "project/group2",
    "project/group3",
    "project/group4"
  ],
  "exp": 1641314954,
  "iat": 1641311354
}
````
It should be replaced with the role-bindings directly
````json
{
  "sub": "julien.chanaud",
  "nbf": 1641311354,
  "roles": [
    "isAdmin()"
  ],
  "iss": "ns4kafka",
  "role-bindings": [
    {
      "namespace": "ns1",
      "verbs": ["GET", "POST", "..."],
      "resources": ["topics", "schemas", "..."]
    },
    {
      "namespace": "ns2",
      "verbs": ["GET", "POST", "..."],
      "resources": ["topics", "schemas", "..."]
    },
  ],
  "exp": 1641314954,
  "iat": 1641311354
}
````

**Comment by piif** [comment]
We currently have the problem : our gitlab server contains many groups, and returned JWT tokens exceed max allowed size, thus micronaut fails to handle it.

In the other hand, role-bindings may change after an API call to apply new ones, thus token should be invalidated after such API calls.

Another idea may be to add the ability to limit groups to scan to a subgroup list specified in a `micronaut.security.gitlab.groups` section ?

**Comment by piif** [comment]
See #362 to limit groups to parent a one given group.
@twobeeb : does this workaround solve your issue ?

**Comment by loicgreffier** [comment]
@piif 

Did you try to increase the header max size on Micronaut as a quick fix ? 

```yml
micronaut:
  server:
    netty:
      max-header-size: 16384
```

https://micronaut-projects.github.io/micronaut-docs-mn2/2.0.3/guide/configurationreference.html#io.micronaut.http.server.netty.configuration.NettyHttpServerConfiguration

A new JWT needs to be generated to take new role bindings/groups in consideration. You can also lower JWT expiration on Micronaut: https://micronaut-projects.github.io/micronaut-security/latest/guide/#jwtGenerator

piif:
Hi @loicgreffier ,
 With a admin account, I had to grow this limit to 32k to make ns4kafka accept the token !
[...]

PR #362, opened 2024-01-29
filter gitlab groups on one parent group

piif:


loicgreffier:
#146 is going to be addressed in #384 to optimize and lighten the content of the JWT. Closing this

PR #383, opened 2024-03-11
Add topic description on Confluent Cloud

ThomasCAI-mlv:


src/main/java/com/michelin/ns4kafka/services/executors/TopicAsyncExecutor.java
```diff
@@ -391,36 +473,62 @@ private void createAndAssociateTags(List<TagTopicInfo> tagsToAssociate) {
...
-                    error -> log.error(String.format("Error dissociating tag %s.",
-                        managedClusterProperties.getConfig().getProperty(CLUSTER_ID) + ":"
-                            + topicName
-                            + "/" + tag), error)));
+                        + ":" + topic.getMetadata().getName(), tag)
+                .subscribe(
+                        success -> {
+                            topic.getMetadata().setGeneration(topic.getMetadata().getGeneration() + 1);
+                            topic.setStatus(Topic.TopicStatus.ofSuccess("Topic tags updated"));
+                            log.info(String.format("Success dissociating tag %s.",
```
loicgreffier, line 520: `topicRepository.create(topic);` missing

src/main/java/com/michelin/ns4kafka/services/executors/TopicAsyncExecutor.java
```diff
@@ -391,36 +473,62 @@ private void createAndAssociateTags(List<TagTopicInfo> tagsToAssociate) {
...
-                            + "/" + tag)),
-                    error -> log.error(String.format("Error dissociating tag %s.",
-                        managedClusterProperties.getConfig().getProperty(CLUSTER_ID) + ":"
-                            + topicName
-                            + "/" + tag), error)));
+                        + ":" + topic.getMetadata().getName(), tag)
+                .subscribe(
+                        success -> {
+                            topic.getMetadata().setGeneration(topic.getMetadata().getGeneration() + 1);
+                            topic.setStatus(Topic.TopicStatus.ofSuccess("Topic tags updated"));
```
loicgreffier, line 519: "Topic tags dissociated"

src/test/java/com/michelin/ns4kafka/services/TopicServiceTest.java
```diff
@@ -984,6 +984,128 @@ void shouldTagsBeInvalidWhenNotConfluentCloud() {
         List<String> validationErrors = topicService.validateTags(ns, topic);
         assertEquals(1, validationErrors.size());
         assertEquals("Invalid value \"TAG_TEST\" for field \"tags\": tags are not currently supported.",
-            validationErrors.get(0));
+                validationErrors.getFirst());
+    }
+
+    @Test
+    void shouldReturnFalseWhenTagFormatIsCorrect() {
```
loicgreffier, line 991: As mentioned before, should return true if format is correct is preferable

src/main/java/com/michelin/ns4kafka/services/TopicService.java
```diff
@@ -318,6 +319,18 @@ public Map<TopicPartition, Long> deleteRecords(Topic topic, Map<TopicPartition,
         }
     }
 
+    /**
+     * Check if topic tags respect confluent format (starts with letter followed by alphanumerical or underscore).
+     *
+     * @param topic     The topic which contains tags
+     * @return Boolean
```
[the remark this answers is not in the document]
ThomasCAI-mlv: Can be turned into positive, indeed

src/main/java/com/michelin/ns4kafka/services/clients/schema/entities/TopicDescriptionUpdateAttributes.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Builder;
+
+/**
+ * Topic description update body's entity's information.
+ *
+ * @param qualifiedName topic entity name
+ * @param description topic description
+ */
+@Builder
+@JsonInclude()
```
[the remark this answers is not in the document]
ThomasCAI-mlv: Ok

src/main/java/com/michelin/ns4kafka/services/executors/TopicAsyncExecutor.java
ThomasCAI-mlv: Right

src/test/java/com/michelin/ns4kafka/services/TopicServiceTest.java
ThomasCAI-mlv: Right

src/main/java/com/michelin/ns4kafka/models/Topic.java
```diff
@@ -64,6 +64,9 @@ public static class TopicSpec {
         @Builder.Default
         @JsonSetter(nulls = Nulls.AS_EMPTY)
         private List<String> tags = new ArrayList<>();
+        @Builder.Default
+        @JsonSetter(nulls = Nulls.AS_EMPTY)
+        private String description = "";
```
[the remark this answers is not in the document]
ThomasCAI-mlv: Not mandatory to set to an empty string, I thought it would be. So ok for the changes.
I also changed the enrichWithDescription to put the null value as default for topic description

src/main/java/com/michelin/ns4kafka/services/executors/TopicAsyncExecutor.java
```diff
@@ -391,36 +473,62 @@ private void createAndAssociateTags(List<TagTopicInfo> tagsToAssociate) {
...
+                        successAssociation -> {
+                            ns4kafkaTopics.stream()
+                                .filter(topic -> tagsToAssociate
+                                    .stream()
+                                    .map(TagTopicInfo::entityName)
+                                    .map(name -> name.substring(name.lastIndexOf(":") + 1))
+                                    .anyMatch(name -> topic.getMetadata().getName().equals(name)))
+                                .forEach(topic -> {
+                                    topic.getMetadata().setGeneration(topic.getMetadata().getGeneration() + 1);
+                                    topic.setStatus(Topic.TopicStatus.ofSuccess("Topic tags updated"));
```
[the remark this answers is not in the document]
ThomasCAI-mlv: Right, and I adjusted the log message to put

src/main/java/com/michelin/ns4kafka/services/clients/schema/entities/TopicListResponseEntityAttributes.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Topics list response's entity's information.
+ *
+ * @param qualifiedName topic entity name
+ * @param description topic description if any
+ * @param name topic name
+ */
+@Builder
+public record TopicListResponseEntityAttributes(String qualifiedName, Optional<String> description, String name) {
```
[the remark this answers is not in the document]
ThomasCAI-mlv: I changed the description from Optional<String> to String and set description to null in topic for cases when there is no description in the get request response or when the topic is not found.
However, I had to get the description from the response outside the stream

src/main/java/com/michelin/ns4kafka/services/executors/TopicAsyncExecutor.java
```diff
@@ -391,36 +473,62 @@ private void createAndAssociateTags(List<TagTopicInfo> tagsToAssociate) {
...
-                                log.info(String.format("Success associating tag %s.", stringTags)),
-                            error ->
-                                log.error(String.format("Error associating tag %s.", stringTags), error)),
+                    schemaRegistryClient.associateTags(managedClusterProperties.getName(), tagsToAssociate)
+                    .subscribe(
+                        successAssociation -> {
+                            ns4kafkaTopics.stream()
+                                .filter(topic -> tagsToAssociate
+                                    .stream()
+                                    .map(TagTopicInfo::entityName)
```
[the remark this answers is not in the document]
ThomasCAI-mlv: I agree, so I built the Mapping withh your suggestion and use this mapping for the loop for status update and logs.

Your assumption seems to be correct, topics without new tags don't need to be updated, so filter can drastically reduce the map size. The filter condition should just be inverted but otherwise, it works

src/main/java/com/michelin/ns4kafka/services/TopicService.java
```diff
@@ -318,6 +319,18 @@ public Map<TopicPartition, Long> deleteRecords(Topic topic, Map<TopicPartition,
         }
     }
 
+    /**
+     * Check if all topic tags respect confluent format (starts with letter followed by alphanumerical or underscore).
+     *
+     * @param topic     The topic which contains tags
+     * @return true if yes, false otherwise
+     */
+    public Boolean isTagsFormatValid(Topic topic) {
```
loicgreffier, line 328: Primitive type `boolean` here

src/main/java/com/michelin/ns4kafka/services/TopicService.java
```diff
@@ -318,6 +319,18 @@ public Map<TopicPartition, Long> deleteRecords(Topic topic, Map<TopicPartition,
...
+        return (topic.getSpec().getTags()
```
loicgreffier, line 329: There's an extra pair of parentheses that is not necessary (I'm suprised neither Sonar or Checkstyle reject it)

src/main/java/com/michelin/ns4kafka/services/TopicService.java
```diff
@@ -338,6 +351,11 @@ public List<String> validateTags(Namespace namespace, Topic topic) {
             return validationErrors;
         }
 
+        if (Boolean.FALSE.equals(isTagsFormatValid(topic))) {
```
loicgreffier, line 354: Simply `if (!isTagsFormatValid(topic))`

src/main/java/com/michelin/ns4kafka/services/executors/TopicAsyncExecutor.java
loicgreffier: OK for that

`topicRepository.create(topic);` is missing to update the topic in the internal topic

src/main/java/com/michelin/ns4kafka/services/clients/schema/entities/TopicDescriptionUpdateResponse.java
```diff
@@ -0,0 +1,15 @@
...
+import org.apache.avro.data.Json;
+
+/**
+ * Update topic description response.
+ *
+ * @param mutatedEntities   The updated entities
+ *
+ */
+@Builder
+public record TopicDescriptionUpdateResponse(Json mutatedEntities) {
```
loicgreffier, line 13: Not sure we should use this pseudo-json type. An actual POJO would be better, but I've seen the signature of the response is wide

Do you actually use the response of the `updateDescription` endpoint ? If no, remove this class for now and just make it return a `Mono<Void>` or `Mono<HttpResponse<Void>>`

src/main/java/com/michelin/ns4kafka/services/executors/TopicAsyncExecutor.java
ThomasCAI-mlv: Right

src/main/java/com/michelin/ns4kafka/services/TopicService.java
ThomasCAI-mlv: OK

src/main/java/com/michelin/ns4kafka/services/TopicService.java
ThomasCAI-mlv: yes

src/main/java/com/michelin/ns4kafka/services/TopicService.java
ThomasCAI-mlv: Sure, I removed them

src/main/java/com/michelin/ns4kafka/models/Topic.java
```diff
@@ -64,6 +64,8 @@ public static class TopicSpec {
         @Builder.Default
         @JsonSetter(nulls = Nulls.AS_EMPTY)
         private List<String> tags = new ArrayList<>();
+        @Builder.Default
+        private String description = null;
```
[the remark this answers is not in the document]
ThomasCAI-mlv: OK

src/main/java/com/michelin/ns4kafka/services/clients/schema/entities/TopicDescriptionUpdateResponse.java
ThomasCAI-mlv: The http response is used for subscription, so it is not possible to have `Mono<Void>`. I tested and `Mono<HttpResponse<Void>>` doesn't work either because the update request returns non null response so the client has trouble to deserialize the response

So I made an actual POJO without the Json type

loicgreffier (approved):

PR #405, opened 2024-06-28
Add documentation about Stream Catalog configuration

ThomasCAI-mlv:


loicgreffier:
@ThomasCAI-mlv Everything's OK for me!

In this section, could we add a word regarding the `config.cluster.id` property that is required to start using the Stream Catalog?

loicgreffier (commented):
@ThomasCAI-mlv Here is my review

PR #410, opened 2024-07-12
Handle wildcard parameter in Topic list API

ThomasCAI-mlv:


ThomasCAI-mlv:
I hesitated to put an "if" in the TopicController list method to handle differently the retrieval of the full list of topics.
The overloaded findAllForNamespace(namespace, param) method can handle that retrieval, but is a bit less optimized than the initial findAllForNamespace(namespace) method.

PR #411, opened 2024-07-15
Handle wildcard parameter in Streams list API

ThomasCAI-mlv:


src/main/java/com/michelin/ns4kafka/service/StreamService.java
```diff
@@ -39,6 +42,21 @@ public List<KafkaStream> findAllForNamespace(Namespace namespace) {
             .toList();
     }
 
+    /**
+     * Find all Kafka Streams of a given namespace, filtered by given parameters.
+     *
+     * @param namespace The namespace
+     * @return A list of Kafka Streams
+     */
+    public List<KafkaStream> findAllForNamespace(Namespace namespace, KafkaStreamFilterParams params) {
```
loicgreffier, line 51: This method is basically the same than the `findByName` method, except it accepts regex special chars. 

I think you can safely update the `findByName` method to accept wildcards in the `String stream` parameter. The method will now return a List<KafkaStream>.

src/main/java/com/michelin/ns4kafka/util/RegexUtils.java
```diff
@@ -0,0 +1,27 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class RegexUtils {
+    /**
+     * Convert wildcard strings list to regex patterns list.
+     *
+     * @param wildcardStrings The list of wildcard strings
+     * @return A list of regex patterns
+     */
+    public static List<String> wildcardStringsToRegexPatterns(List<String> wildcardStrings) {
+        return wildcardStrings.stream()
```
loicgreffier, line 19: Could you add unit tests for this class? I think we should verify:
- For a given input, the output matches what it is expected.
- For the output regex, it matches to expected strings.

I think the following tests should be a good start:
- abc.my* / abc.myTopic ➡️ Yes
- abc.my?opic / abc.myTopic ➡️ Yes
- abc.my*opics / abc.myTopic ➡️ No, because 's' in the left value
- abc.my*opic / abc.myTopics ➡️ No, because 's' in the right value
- abc.my*opic / abc.myAbcdefghiTopic ➡️ Yes
- abc.myT*op?c / abc.myTopiiic ➡️ No, too much 'i'
- abc.myT*o?ic / abc.myTopic ➡️ Yes
- abc.myT*o*ic / abc.myTopic ➡️ Yes
- abc.my*?p*c / abc.myTopic ➡️ Yes
- abc.myTopic** / abc.myTopic ➡️ Yes

I think handling `*` (0 or n) and `?` (0 or 1) is a good first step.

loicgreffier (commented):
@ThomasCAI-mlv Here is my first review

src/main/java/com/michelin/ns4kafka/service/StreamService.java
ThomasCAI-mlv: I think the findByName is used for the endpoint /streams/{stream} which is then used by old clients.

src/main/java/com/michelin/ns4kafka/util/RegexUtils.java
ThomasCAI-mlv: I think `?` is commonly used to match only 1 character (not 0)
Examples for microsoft: https://support.microsoft.com/en-us/office/examples-of-wildcard-characters-939e153f-bd30-47e4-a763-61897c87b3f4

src/main/java/com/michelin/ns4kafka/controller/StreamController.java
```diff
@@ -41,9 +42,12 @@ public class StreamController extends NamespacedResourceController {
      * @param namespace The namespace
      * @return A list of Kafka Streams
      */
-    @Get("/")
-    List<KafkaStream> list(String namespace) {
-        return streamService.findAllForNamespace(getNamespace(namespace));
+    @Get
+    List<KafkaStream> list(String namespace, @QueryValue Optional<List<String>> name) {
```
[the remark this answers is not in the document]
ThomasCAI-mlv: All comments taken into account

src/main/java/com/michelin/ns4kafka/util/RegexUtils.java
ThomasCAI-mlv: I added the filterByPattern method in RegexUtils to filter any string with regex patterns.
So we have the filter by wildcard by combining:
1) the conversion from wildcard to regex
2) the filter by regex pattern

I added:
- TU for filter from regex patterns
- functional tests for filter from wildcards

src/main/java/com/michelin/ns4kafka/service/StreamService.java
ThomasCAI-mlv: As seen together:
- I'll keep the findByName implementation since we prefer that old clients use the endpoint `/streams/{name}` which uses the `findByName` method and wildcard will not be supported for them: any wildcard will result in "not found".
- I'll rename the `findAllForNamespace` method to `findByWildcardName` which accepts wildcard names, used for filter.
- I'll put the `@Deprecated` annotation in the controller method corresponding to the `/streams/{name}` endpoint.

loicgreffier:
@ThomasCAI-mlv LGTM

PR #422, opened 2024-07-30
Fix connect cluster apply when there is no connect cluster in Ns4Kafka configuration

ThomasCAI-mlv:


loicgreffier:
@ThomasCAI-mlv 

LGTM, I've pushed a commit with some naming convention updates. Not directly related to your fix

PR #427, opened 2024-08-06
Handle wildcard parameter in ResourceQuota list API

ThomasCAI-mlv:


ThomasCAI-mlv:
PR to be iso with other GET endpoints
Quota objects are unique per namespace, so the wildcard search of quota has little plus-value

src/main/java/com/michelin/ns4kafka/controller/quota/ResourceQuotaController.java
```diff
@@ -43,9 +43,11 @@ public class ResourceQuotaController extends NamespacedResourceController {
      * @return A list of quotas
      */
     @Get
-    public List<ResourceQuotaResponse> list(String namespace) {
-        return List.of(resourceQuotaService.getUsedResourcesByQuotaByNamespace(getNamespace(namespace),
-            resourceQuotaService.findByNamespace(namespace)));
+    public List<ResourceQuotaResponse> list(String namespace, @QueryValue(defaultValue = "*") String name) {
+        return resourceQuotaService.findByWildcardName(namespace, name)
+            .stream()
+            .map(o -> resourceQuotaService.getUsedResourcesByQuotaByNamespace(getNamespace(namespace), Optional.of(o)))
```
loicgreffier, line 49: `resourceQuota` will suit better than `o`

src/main/java/com/michelin/ns4kafka/service/ResourceQuotaService.java
```diff
@@ -54,15 +55,43 @@ public class ResourceQuotaService {
...
         return resourceQuotaRepository.findForNamespace(namespace);
     }
 
+    /**
+     * Find all the resource quota of a given namespace.
+     *
+     * @param namespace The namespace used to research
+     * @return The researched resource quota
+     */
+    public List<ResourceQuota> findAllForNamespace(String namespace) {
```
loicgreffier, line 73: Literally the same method than `resourceQuotaRepository#findForNamespace`, right ?

```java
getKafkaStore()
  .values()
  .stream()
  .filter(resourceQuota -> resourceQuota.getMetadata().getNamespace().equals(namespace))
```

As it is used only from `resourceQuotaService#findByWildcardName`, you can just do:

```java
findForNamespace(namespace)
  .stream()
  .filter(quota -> RegexUtils.filterByPattern(quota.getMetadata().getName(), nameFilterPatterns))
  .toList();
```

Optional can be converted to stream: https://javadoc.scijava.org/Java13/java.base/java/util/Optional.html#stream()

loicgreffier (commented):
@ThomasCAI-mlv Minor updates before merging

src/main/java/com/michelin/ns4kafka/controller/quota/ResourceQuotaController.java
ThomasCAI-mlv: Taken into account

PR #428, opened 2024-08-12
Delete a specific version of a schema

ThomasCAI-mlv:


src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
```diff
@@ -142,16 +142,18 @@ public Mono<HttpResponse<Schema>> apply(String namespace, @Valid @Body Schema sc
...
 
     /**
-     * Delete all schemas under the given subject.
+     * Delete all schema versions under the given subject, or a specific version of the schema if specified.
      *
-     * @param namespace The current namespace
-     * @param subject   The current subject to delete
+     * @param namespace The namespace
+     * @param subject   The subject
+     * @param version   The version of the schema to delete (optional)
```
loicgreffier, line 149: No need to set `(optional)` here

src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
```diff
@@ -142,16 +142,18 @@ public Mono<HttpResponse<Schema>> apply(String namespace, @Valid @Body Schema sc
...
      * @param dryrun    Run in dry mode or not
      * @return A HTTP response
      */
     @Status(HttpStatus.NO_CONTENT)
     @Delete("/{subject}")
     public Mono<HttpResponse<Void>> delete(String namespace, @PathVariable String subject,
```
loicgreffier, line 155: Could you indent like:

```java
String namespace, 
@PathVariable String subject,
@QueryValue Optional<String> version,
@QueryValue(defaultValue = "false") boolean dryrun
```

src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
```diff
@@ -160,24 +162,59 @@ public Mono<HttpResponse<Void>> delete(String namespace, @PathVariable String su
             return Mono.error(new ResourceValidationException(SCHEMA, subject, invalidOwner(subject)));
         }
 
-        return schemaService.getLatestSubject(ns, subject)
+        // delete all versions of the schema
+        if (version.isEmpty()) {
```
loicgreffier, line 166: OK for this `if-else` logic. It looks it can be simplified in a single expression like:

```java
return version
  .<Mono<HttpResponse<Void>>>map(v -> schemaService.getSubject(ns, subject, v)
    ...
  .orElseGet(() -> schemaService.getLatestSubject(ns, subject)
    ...
  );
```

src/main/java/com/michelin/ns4kafka/service/client/schema/SchemaRegistryClient.java
```diff
@@ -137,6 +138,25 @@ public Mono<Integer[]> deleteSubject(String kafkaCluster, String subject, boolea
         return Mono.from(httpClient.retrieve(request, Integer[].class));
     }
 
+    /**
+     * Delete a subject.
+     *
+     * @param kafkaCluster The Kafka cluster
+     * @param subject      The subject
+     * @param version       The version
```
loicgreffier, line 146: indentation

src/test/java/com/michelin/ns4kafka/controller/SchemaControllerTest.java
```diff
@@ -442,15 +443,38 @@ void shouldNotUpdateCompatibilityWhenNamespaceNotOwner() {
     }
 
     @Test
-    void shouldNotDeleteSchemaWhenNotOwner() {
+    void shouldNotDeleteAllSchemaVersionsWhenNotOwner() {
+        Namespace namespace = buildNamespace();
+
+        when(namespaceService.findByName("myNamespace"))
+                .thenReturn(Optional.of(namespace));
```
loicgreffier, line 450: Indentation

src/test/java/com/michelin/ns4kafka/controller/SchemaControllerTest.java
```diff
@@ -474,23 +499,126 @@ void shouldDeleteSchema() {
...
+            .thenReturn(false);
+        doNothing().when(applicationEventPublisher).publishEvent(any());
+
+        StepVerifier.create(schemaController.delete("myNamespace", "prefix.subject-value",
+            Optional.of("latest"), false))
             .consumeNextWith(response -> assertEquals(HttpStatus.NO_CONTENT, response.getStatus()))
             .verifyComplete();
 
-        verify(schemaService).delete(namespace, "prefix.subject-value");
+        verify(applicationEventPublisher, times(1)).publishEvent(any());
```
loicgreffier, line 564: You can remove `times(1)` and the associated import, this is the default value used by `verify`

src/test/java/com/michelin/ns4kafka/controller/SchemaControllerTest.java
```diff
@@ -474,23 +499,126 @@ void shouldDeleteSchema() {
...
+            .thenReturn(Optional.of("test-user"));
+        when(securityService.hasRole(ResourceBasedSecurityRule.IS_ADMIN))
+            .thenReturn(false);
+        doNothing().when(applicationEventPublisher).publishEvent(any());
+
+        StepVerifier.create(schemaController.delete("myNamespace", "prefix.subject-value", Optional.of("2"), false))
+            .consumeNextWith(response -> assertEquals(HttpStatus.NO_CONTENT, response.getStatus()))
+            .verifyComplete();
+
+        verify(applicationEventPublisher, times(1)).publishEvent(any());
```
loicgreffier, line 591: Same than above for `times(1)`

src/test/java/com/michelin/ns4kafka/controller/SchemaControllerTest.java
```diff
@@ -474,23 +499,126 @@ void shouldDeleteSchema() {
...
+            .thenReturn(Optional.of("test-user"));
+        when(securityService.hasRole(ResourceBasedSecurityRule.IS_ADMIN))
+            .thenReturn(false);
+        doNothing().when(applicationEventPublisher).publishEvent(any());
+
+        StepVerifier.create(schemaController.delete("myNamespace", "prefix.subject-value", Optional.of("1"), false))
+            .consumeNextWith(response -> assertEquals(HttpStatus.NO_CONTENT, response.getStatus()))
+            .verifyComplete();
+
+        verify(applicationEventPublisher, times(1)).publishEvent(any());
```
loicgreffier, line 617: Same than above for `times(1)`

src/test/java/com/michelin/ns4kafka/integration/SchemaIntegrationTest.java
```diff
@@ -766,5 +766,53 @@ void shouldRegisterSameSchemaTwice() {
         assertNotNull(schemaAfterPostOnRegistry.id());
         assertEquals(2, schemaAfterPostOnRegistry.version());
         assertEquals("ns1-subject3-value", schemaAfterPostOnRegistry.subject());
+
+        // Delete latest schema version
```
loicgreffier, line 770: I think a dedicated integration test for deletion would be preferable:
- Register 4 schema versions
- Delete "latest" (v4). Check all schemas response
- Delete old version "v2". Check all schemas response
- Delete all (no version param). Check all schemas response
What do you think?

loicgreffier (commented):
@ThomasCAI-mlv 

Here is my review. Mostly minor adjustments and a dedicated integration test for all the deletion use-cases.

src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
ThomasCAI-mlv: OK

src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
ThomasCAI-mlv: OK

src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
ThomasCAI-mlv: The actions performed between the two cases are not the same:

1. In the first case `version.isEmpty()` (need to delete all versions), we just call the `deleteAllVersions` method to delete all the versions of the schema.
2. In the second case (need to delete a specific version), we call the `deleteVersion` method to delete a specific version, and depending on the situation, we need to send different event log:
- If we deleted the **last** version of the schema, we send a `deleted` status. (no schema is remaining)
- If we deleted the **latest** version but there is still one remaining version, we send a `changed` status.
- If we deleted the non-latest version, we send no event log. (the latest version was not deleted and remains the latest version)

loicgreffier: Agree with the whole logic, I was just suggesting to update the `if-else` condition with the `Optional#orElseGet` logic

src/main/java/com/michelin/ns4kafka/service/client/schema/SchemaRegistryClient.java
ThomasCAI-mlv: OK

src/test/java/com/michelin/ns4kafka/controller/SchemaControllerTest.java
ThomasCAI-mlv: OK

src/test/java/com/michelin/ns4kafka/controller/SchemaControllerTest.java
ThomasCAI-mlv: OK

src/test/java/com/michelin/ns4kafka/controller/SchemaControllerTest.java
ThomasCAI-mlv: OK

src/test/java/com/michelin/ns4kafka/controller/SchemaControllerTest.java
ThomasCAI-mlv: OK

src/test/java/com/michelin/ns4kafka/integration/SchemaIntegrationTest.java
ThomasCAI-mlv: I hesitated between having a dedicated tests or enriching the existing test. I just did the one that took less effort. I can write a dedicated one

src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
```diff
@@ -160,24 +163,59 @@ public Mono<HttpResponse<Void>> delete(String namespace, @PathVariable String su
...
+                    .flatMap(optionalNewLatestSubject -> {
+                        // there was only one version & it was deleted, so the schema is deleted in ns4
+                        if (optionalNewLatestSubject.isEmpty()) {
+                            sendEventLog(schemaToDelete, ApplyStatus.deleted,
+                                schemaToDelete.getSpec(), null);
+                        }
+                        // there were multiple versions & latest version was deleted, so the schema is changed in ns4
+                        if (optionalNewLatestSubject.isPresent() && optionalNewLatestSubject.get()
+                            .getSpec().getVersion() < schemaToDelete.getSpec().getVersion()) {
+                            sendEventLog(schemaToDelete, ApplyStatus.changed,
```
loicgreffier, line 214: Always send a `deleted` status, maybe specifying the version additionally

ThomasCAI-mlv: OK, just with the `deleted` status, I refactored the controller method.

src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
ThomasCAI-mlv: By changing the sentEventLog to sending only `deleted` status, I setup version.map().orElseGet()

src/main/java/com/michelin/ns4kafka/controller/SchemaController.java
ThomasCAI-mlv: So I updated the sendEventLog method to add a `version` argument, so I updated the method sendEventLog in many other places too

PR #430, opened 2024-08-19
partially solve kafkactl issue 91

piif:
Implements APIs entrypoints to set and check a user password :
 - `PATCH /api/namespaces/[namespace]/users/[user]/set-password` with new password as body ; returns same result than reset-password (open to discussion : it may just return a status ?)
 - `POST /api/namespaces/[namespace]/users/[user]/check-password` with new password as body ; return 200 is password matches current one, 401 else

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/controller/UserController.java
+++ b/src/main/java/com/michelin/ns4kafka/controller/UserController.java
@@ -12,10 +12,13 @@ import com.michelin.ns4kafka.util.enumation.ApplyStatus;
 import com.michelin.ns4kafka.util.exception.ResourceValidationException;
 import io.micronaut.context.ApplicationContext;
 import io.micronaut.http.HttpResponse;
+import io.micronaut.http.annotation.Body;
 import io.micronaut.http.annotation.Controller;
 import io.micronaut.http.annotation.Post;
+import io.micronaut.http.annotation.Patch;
 import io.micronaut.inject.qualifiers.Qualifiers;
 import io.swagger.v3.oas.annotations.tags.Tag;
+import lombok.extern.slf4j.Slf4j;
 import jakarta.inject.Inject;
 import java.time.Instant;
 import java.util.Date;
@@ -23,6 +26,7 @@ import java.util.Date;
 /**
  * Controller to manage users.
  */
+@Slf4j
 @Tag(name = "Users", description = "Manage the users.")
 @Controller(value = "/api/namespaces/{namespace}/users")
 public class UserController extends NamespacedResourceController {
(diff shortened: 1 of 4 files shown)


src/main/java/com/michelin/ns4kafka/controller/UserController.java
```diff
@@ -64,4 +98,30 @@ public HttpResponse<KafkaUserResetPassword> resetPassword(String namespace, Stri
...
+        Namespace ns = getNamespace(namespace);
+
+        if (!ns.getSpec().getKafkaUser().equals(user)) {
+            throw new ResourceValidationException(KAFKA_USER_CHECK_PASSWORD, user, invalidKafkaUser(user));
+        }
+
+        UserAsyncExecutor userAsyncExecutor =
+            applicationContext.getBean(UserAsyncExecutor.class, Qualifiers.byName(ns.getMetadata().getCluster()));
+
+        if (userAsyncExecutor.checkPassword(ns.getSpec().getKafkaUser(), password)) {
```
loicgreffier, line 122: As suggested in https://github.com/michelin/kafkactl/issues/91, a status that specifies if the connection with the given password ends with `SUCCESS` or `FAIL` may be more significant.

src/main/java/com/michelin/ns4kafka/service/executor/UserAsyncExecutor.java
```diff
@@ -200,6 +233,66 @@ public String resetPassword(String user) {
             return password;
         }
 
+        @Override
+        public void setPassword(String user, String password) {
```
loicgreffier, line 237: See suggestions from above:

➡️ I think `userAsyncExecutor#resetPassword` and `userAsyncExecutor#setPassword` can be a single method as well:

- Add a password parameter to userAsyncExecutor#resetPassword
- Extract the generation of the random password outside of `userAsyncExecutor#resetPassword` in order to generate it if the @Body is empty

src/main/java/com/michelin/ns4kafka/service/executor/UserAsyncExecutor.java
```diff
@@ -200,6 +233,66 @@ public String resetPassword(String user) {
...
+                Thread.currentThread().interrupt();
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        @Override
+        public boolean checkPassword(String user, String password, Properties config) {
+            Properties props = new Properties();
+            props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, config.get("bootstrap.servers"));
```
loicgreffier, line 253: Using an `AdminClient` to verify the connexion may be more elegant than instantiating a dedicated producer, let me sleep on that

loicgreffier (commented):
@piif Here is my review. Overall:

- `resetPassword` and `setPassword` could be a single endpoint
- `checkPassword` could rely on AdminClient

PR #443, opened 2024-09-18
Handle wildcard parameter in Topic deletion API

ThomasCAI-mlv:


src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -203,6 +203,25 @@ public void deleteTopic(Topic topic) throws InterruptedException, ExecutionExcep
             managedClusterProperties.getName());
     }
 
+    /**
+     * Delete a list of topics.
+     *
+     * @param topics The topics to delete
+     */
+    public void deleteTopics(List<Topic> topics) throws InterruptedException, ExecutionException, TimeoutException {
```
loicgreffier, line 211: Little improvement:
https://github.com/michelin/ns4kafka/blob/ae3e5d9a145a6869c304a56d5dbda0bbb5f4ea37/src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java#L197 can call `deleteTopics(List<Topics> topics)` like

public void deleteTopic(Topic topic) {
  deleteTopics(List.of(topic));
}

loicgreffier (commented):
@ThomasCAI-mlv Just one improvement, otherwise LGTM

src/main/java/com/michelin/ns4kafka/controller/topic/TopicController.java
```diff
@@ -156,16 +156,55 @@ public HttpResponse<Topic> apply(String namespace, @Valid @Body Topic topic,
...
+
     /**
      * Delete a topic.
      *
      * @param namespace The namespace
      * @param topic     The topic
-     * @param dryrun    Is dry run mode or not ?
+     * @param dryrun    Is dry run mode or not?
      * @return An HTTP response
+     * @deprecated use bulkDelete(String, String name, boolean) instead.
```
loicgreffier, line 203: `{@link #bulkDelete(String, String, boolean)}` better 😉

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
ThomasCAI-mlv: Maybe, it is even better to completely remove the method and replace it by `deleteTopics(List<Topic> topics)`.
What do you think?

PR #446, opened 2024-09-25
Handle wildcard parameter in Schema deletion API

ThomasCAI-mlv:


loicgreffier:
@ThomasCAI-mlv Linking https://github.com/michelin/ns4kafka/pull/447. Note that `notFound` is thrown when version does not match among all the schemas

ThomasCAI-mlv:
I adopted your errorHandling structure for "notFound" and "dryRun" cases.

As discussed together in case of deletion of version:
1) I removed the ownership checker, as we don't want to restrict the usage of wildcard by forcing users to type the namespace prefix. The findByWildcard already filters the namespace's resources.

2) We prefer anyMatch over allMatch to be homogeneous with other DELETE endpoint:
- Deleting multiple connect clusters including at least one with connectors will lead to "ResourceValidationError"
- Deleting multiple namespaces including at least one with resources inside will lead to "ResourceValidationError"
So we can also implement the schema version deletion and throw "notFound" if at least one version doesn't exist.

loicgreffier (approved):
LGTM

PR #466, opened 2024-10-11
Fix return type for Schema list API

ThomasCAI-mlv:
The `list` schema endpoint with wildcard will be the main endpoint to get one or multiple schemas.
This PR handles whether the returned information from this endpoint is filled with compatibility, schema id, type, version or not:
- When there is one schema returned, the endpoint returns all information.
- When there are 2 schemas or more, the endpoint only returns the subject name.


loicgreffier (approved):
LGTM

PR #469, opened 2024-10-15
Fix connect worker getting rebalanced in integration tests

loicgreffier:


ThomasCAI-mlv (commented):
OK for me. Should we leave the sleep at 4 sec instead of the initial 3 sec you put in the connector integration tests?

PR #491, opened 2024-12-05
Find namespace from topic name

ThomasCAI-mlv:
Add the possibility to find the namespace which is the owner of a topic, thanks to a given topic name.

The PR adds the `topic` query parameter in the LIST namespace API:
- if the query parameter is an empty string, the endpoint doesn't filter at all by the topic name.
- if the query parameter is not empty, the endpoint finds the namespace which is the owner of the topic. It returns no namespace if no namespace is owner of the topic (or the topic doesn't exist).

Maybe, the implementation could be 2 additional parameters, resource type and resource name, to allow the search on all types of resources (not only topics), I left it to discussion.


loicgreffier (approved):
@ThomasCAI-mlv OK for me

PR #494, opened 2024-12-11
Improve connect cluster vault error handling

ThomasCAI-mlv:
For the connect cluster vault API, I simplified the `validateConnectClusterVault` method check:
- If the provided connect cluster is not found (the connect cluster does not exist or the namespace does not have the OWNER or WRITE ACL on this connect cluster), throw an "not found" error
- If the provided connect cluster is found but is missing AES config fields, throw an "no config" error. I updated the error message to be more explicit, because previously, it was confusing to display the list of other usable vault connect clusters, since the user explicitly gave the connect cluster he wants to use to vault his secrets. In any case, they can use the LIST vaults API to list the available vaults connect clusters.

I also removed the "allowed namespace" check because it was weird to provide information of another namespaces to a given namespace.

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
+++ b/src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
@@ -1,7 +1,6 @@
 package com.michelin.ns4kafka.controller.connect;
 
 import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidConnectClusterDeleteOperation;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidConnectClusterNotAllowed;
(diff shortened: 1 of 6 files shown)


loicgreffier (approved):
@ThomasCAI-mlv Good improvements 👍

- I updated the error message `AES key and salt are required to use connect-cluster vault` to not have a command-line on server side 
- I had some doubts about removing: https://github.com/michelin/ns4kafka/blob/95f399b61a4ed60e2663ed6edeb7e6de3246160b/src/main/java/com/michelin/ns4kafka/service/ConnectClusterService.java#L318 For connectors, if there is an error on the `connectCluster`, the list is suggested: https://github.com/michelin/ns4kafka/blob/95f399b61a4ed60e2663ed6edeb7e6de3246160b/src/main/java/com/michelin/ns4kafka/service/ConnectorService.java#L132
What's your opinion?

PR #507, opened 2025-01-31
Improve catalog information synchronisation with GraphQL

ThomasCAI-mlv:
**1. Context**

- Currently, we have the possibility to add tags and description to a topic and ns4kafka synchronizes these information with **Confluent Cloud** catalog information (tags & description). This synchronisation is performed with the Stream Catalog REST API ([Stream Catalog documentation](https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html)).

- However, this API is not suited for the synchronisation because of the **500 topics limit** per call, and the fact that **there is no filter** allowing to **query only topics with a description** (though it is possible to query only topics with at least one tag). These drawbacks lead to having to query all the cluster topics 500 per 500 in order to perform the synchronization, which is quite bad for performance.

**2. Proposed implementation**
This PR adds GraphQL API calls to query the topic lists with their tags and description. This API overcomes the two previous problems: it is possible to query the list of topics with a description, and query the list of topics with at least one tag, without pagination.

This API still has drawbacks:
- it is only usable with the advanced Stream Governance Package on Confluent Cloud, which is the non-free package at 1$/hour: so we first try to use GraphQL API, and if it fails, we use the old Stream Catalog REST API to get the topics.
- In order to query topics with at least one tag, all the existing tags need to be filled in the GraphQL query, so another API call to Stream Catalog shall be made to get the tags list.
- I didn't find a way to query the list of topics with a description **OR** with at least one tag. I had to split the call: so one GraphQL API call to get the topics with a description, another one to get the topics with at least one tag. If you want to challenge it, you can try the GraphQL API (doc: https://docs.confluent.io/cloud/current/stream-governance/graphql-apis.html#definition-kafka_topic)

**3. Other**
- I added a `sync-catalog` property which must be set to `true` in order to allow the synchronization of catalog information in Confluent Cloud. It is set at `false` by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,18 +295,23 @@ The configuration will depend on the authentication method selected for your bro
 
 ### Stream Catalog
 
-For Confluent Cloud only, the Stream Catalog API is used to retrieve the list of topics with their tags and their description. Reminder that the `config.cluster.id` parameter from [managed cluster properties](#managed-clusters) must be set to use Confluent Cloud.
+For Confluent Cloud only, you can activate the synchronization of topic tags and description in ns4kafka with topic tags and description in Confluent Cloud.
 
-You can configure the page size of the response of the API using the following properties:
+The synchronization is done with the Confluent Stream Catalog GraphQL API if you have the appropriate Stream package on Confluent, otherwise with the Confluent Stream Catalog REST API.
+
+You can configure the synchronization using the following properties:
 
 ```yaml
 ns4kafka:
   confluent-cloud:
     stream-catalog:
       page-size: 500
+      sync-catalog: true
 ```
 
-The max page size is at 500 as described in the [Confluent Cloud documentation](https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html#list-all-topics).
+The max page size is used for the Stream Catalog API and is capped at 500 as described in the [Confluent Cloud documentation](https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html#list-all-topics).
+
+Reminder that the `config.cluster.id` parameter from [managed cluster properties](#managed-clusters) must be set to use Confluent Cloud.
 
 ### Managed Kafka Clusters
 
--- a/README.md
+++ b/README.md
@@ -295,9 +295,9 @@ The configuration will depend on the authentication method selected for your bro
 
 ### Stream Catalog
 
-For Confluent Cloud only, you can activate the synchronization of topic tags and description in ns4kafka with topic tags and description in Confluent Cloud.
+For Confluent Cloud only, topic tags and description can be synchronized with Ns4kafka.
 
-The synchronization is done with the Confluent Stream Catalog GraphQL API if you have the appropriate Stream package on Confluent, otherwise with the Confluent Stream Catalog REST API.
+The synchronization is done with the [Confluent Stream Catalog GraphQL API](https://docs.confluent.io/cloud/current/stream-governance/graphql-apis.html) if you have the appropriate Stream Governance package on Confluent, otherwise with the [Confluent Stream Catalog REST API](https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html#list-all-topics).
 
 You can configure the synchronization using the following properties:
 
@@ -309,7 +309,7 @@ ns4kafka:
       sync-catalog: true
 ```
 
-The max page size is used for the Stream Catalog API and is capped at 500 as described in the [Confluent Cloud documentation](https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html#list-all-topics).
+The page size is used for the Stream Catalog REST API and is capped at 500 as described in the [Confluent Cloud documentation](https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html#limits-on-topic-listings).
 
 Reminder that the `config.cluster.id` parameter from [managed cluster properties](#managed-clusters) must be set to use Confluent Cloud.
 
--- a/src/main/java/com/michelin/ns4kafka/service/client/schema/SchemaRegistryClient.java
+++ b/src/main/java/com/michelin/ns4kafka/service/client/schema/SchemaRegistryClient.java
@@ -20,6 +20,7 @@
 package com.michelin.ns4kafka.service.client.schema;
 
 import com.michelin.ns4kafka.property.ManagedClusterProperties;
+import com.michelin.ns4kafka.service.client.schema.entities.GraphQueryResponse;
 import com.michelin.ns4kafka.service.client.schema.entities.SchemaCompatibilityCheckResponse;
 import com.michelin.ns4kafka.service.client.schema.entities.SchemaCompatibilityRequest;
 import com.michelin.ns4kafka.service.client.schema.entities.SchemaCompatibilityResponse;
@@ -47,6 +48,7 @@ import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import lombok.extern.slf4j.Slf4j;
 import reactor.core.publisher.Flux;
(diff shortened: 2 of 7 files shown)


src/main/java/com/michelin/ns4kafka/property/ConfluentCloudProperties.java
```diff
@@ -40,5 +40,6 @@ public class ConfluentCloudProperties {
     @ConfigurationProperties("stream-catalog")
     public static class StreamCatalogProperties {
         private int pageSize = 500;
+        private boolean syncCatalog;
```
loicgreffier, line 43: Should be added to the documentation

src/main/java/com/michelin/ns4kafka/service/client/schema/SchemaRegistryClient.java
```diff
@@ -354,11 +373,53 @@ public Mono<TopicListResponse> getTopicWithCatalogInfo(String kafkaCluster, int
...
     }
 
+    /**
+     * Query Stream Catalog information, using GraphQL.
+     *
+     * @param kafkaCluster The Kafka cluster
+     * @param query        The GraphQL query
+     * @return The GraphQL response
+     */
+    public Mono<GraphQueryResponse> queryWithGraphQl(String kafkaCluster, String query) {
```
loicgreffier, line 383: Can this function be private?

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -294,40 +307,91 @@ public void alterDescriptions(List<Topic> ns4kafkaTopics, Map<String, Topic> bro
...
-                            .setTags(topicEntity.classificationNames());
-                        topics.get(topicEntity.attributes().name()).getSpec()
-                            .setDescription(topicEntity.attributes().description());
-                    });
-                offset += limit;
-            } while (!topicListResponse.entities().isEmpty());
+        if (confluentCloudProperties.getStreamCatalog().isSyncCatalog()
+            && managedClusterProperties.isConfluentCloud()) {
+            try {
+                enrichWithCatalogInfoWithGraphQl(topics);
```
loicgreffier, line 319: Rename to `enrichWithGraphQlCatalogInfo`?

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -294,40 +307,91 @@ public void alterDescriptions(List<Topic> ns4kafkaTopics, Map<String, Topic> bro
...
+            } catch (Exception e) {
+                enrichWithCatalogInfoWithStreamCatalog(topics);
```
loicgreffier, line 321: Rename to `enrichWithCatalogInfo`? (which suggest it is the default way to enrich with catalog info)

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -294,40 +307,91 @@ public void alterDescriptions(List<Topic> ns4kafkaTopics, Map<String, Topic> bro
...
         }
     }
 
+    /**
+     * Enrich topics with Confluent catalog information, using Confluent GraphQL API.
+     *
+     * @param topics Topics to enrich
+     */
+    public void enrichWithCatalogInfoWithGraphQl(Map<String, Topic> topics) {
+        schemaRegistryClient.getTopicsWithDescriptionWithGraphQl(managedClusterProperties.getName())
```
loicgreffier, line 332: There's no way to request both description and tags at the same time?

loicgreffier (commented):
@ThomasCAI-mlv Here is my first review

src/main/java/com/michelin/ns4kafka/property/ConfluentCloudProperties.java
ThomasCAI-mlv: Added

src/main/java/com/michelin/ns4kafka/service/client/schema/SchemaRegistryClient.java
ThomasCAI-mlv: yes, I changed it

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
ThomasCAI-mlv: Renamed

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
ThomasCAI-mlv: I renamed it `enrichWithRestCatalogInfo` to highlight the difference on the API type (REST vs GraphQL). What do you think?

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
ThomasCAI-mlv: Unfortunately I don't think it is possible with the way Confluent manages their tags entities.

The query to get the topics with description uses the default keyword `where`:
`query {
	kafka_topic(where: {description: {_gte: null}}
{
    nameLower
		tags
	}
}`

The query to get the topics with tags uses the keyword `tags`:

`query {
	kafka_topic(tags: ["D3", "API_CATALOG", "TAG1", "TAG2", "TAG3", "SNW_CONNECT_CLUSTER", "CMDB", "POV", "MAINTENANCE", "FEEDBACK"])
	{
    nameLower
		tags
	}
}`

Merging both keywords leads to query on topics with a description AND with tags. But we want an OR logic...

If we had the tags condition in the `where` keyword, it would be possible.

README.md
```diff
@@ -295,18 +295,23 @@ The configuration will depend on the authentication method selected for your bro
 
 ### Stream Catalog
 
-For Confluent Cloud only, the Stream Catalog API is used to retrieve the list of topics with their tags and their description. Reminder that the `config.cluster.id` parameter from [managed cluster properties](#managed-clusters) must be set to use Confluent Cloud.
+For Confluent Cloud only, you can activate the synchronization of topic tags and description in ns4kafka with topic tags and description in Confluent Cloud.
 
-You can configure the page size of the response of the API using the following properties:
+The synchronization is done with the Confluent Stream Catalog GraphQL API if you have the appropriate Stream package on Confluent, otherwise with the Confluent Stream Catalog REST API.
```
loicgreffier, line 300: It can be cool to add the link like:

```markdown
... is done using the [Confluent Stream Catalog GraphQL API](https://docs.confluent.io/cloud/current/stream-governance/graphql-apis.html)
```

README.md
```diff
@@ -295,18 +295,23 @@ The configuration will depend on the authentication method selected for your bro
...
 ```yaml
 ns4kafka:
   confluent-cloud:
     stream-catalog:
       page-size: 500
+      sync-catalog: true
 ```
 
-The max page size is at 500 as described in the [Confluent Cloud documentation](https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html#list-all-topics).
+The max page size is used for the Stream Catalog API and is capped at 500 as described in the [Confluent Cloud documentation](https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html#list-all-topics).
```
loicgreffier, line 312: Did you mean for the Stream Catalog **REST** API?

loicgreffier, line 312: Some suggestions:
- Start with "The maximum `page-size`"
- Precise Stream Catalog **REST** API

loicgreffier (commented):
@ThomasCAI-mlv Some suggestions on the doc, otherwise it can be merged

README.md
ThomasCAI-mlv: Yes, I added the link for the REST API too.
In fact, I hesitated between these 2 pages:
- Explanation page of the usage: https://docs.confluent.io/cloud/current/stream-governance/stream-catalog-rest-apis.html#list-all-topics
- API specs: https://docs.confluent.io/cloud/current/api.html#tag/Search-(v1)/operation/searchUsingBasic

I took the first one because it is difficult to understand how the endpoint is related to topic list query with the specs.

README.md
ThomasCAI-mlv: OK

PR #513, opened 2025-02-13
Fix ignored default Stream catalog page size

loicgreffier:


loicgreffier:
@ThomasCAI-mlv 

When 

```yml
ns4kafka:
  confluent-cloud:
    stream-catalog:
```

is not explicitely defined in the `application.yml`, default value is ignored and Ns4Kafka fails:

```console
java.lang.NullPointerException: Cannot invoke "com.michelin.ns4kafka.property.ConfluentCloudProperties$StreamCatalogProperties.getPageSize()" because the return value of "com.michelin.ns4kafka.property.ConfluentCloudProperties.getStreamCatalog()" is null
```

ThomasCAI-mlv (approved):
OK

PR #516, opened 2025-02-16
Fix inconsistent failure in Connector integration tests due to unassigned tasks

loicgreffier:


loicgreffier:
@ThomasCAI-mlv As a workaround for inconsistent failures, I suggest to not check the task's state but only the connector state as some tasks remain unassigned after being deployed to Kafka Connect.

In addition, the PR brings some fixes and improvements:

- `CONNECT_PRODUCER` and `CONNECT_CONSUMER` configs were missing on Kafka Connect causing connectors to fail authenticating to broker.
- Connectors and topics have been renamed. Topics are not shared anymore across connectors to avoid confusion.

ThomasCAI-mlv (approved):

PR #577, opened 2025-06-11
Automatically import changelog and repartition topics

loicgreffier:


src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -177,10 +186,65 @@ public void synchronizeTopics() {
...
+                    topic.getMetadata().setCreationTimestamp(Date.from(Instant.now()));
+                    topic.getMetadata().setCluster(managedClusterProperties.getName());
+                    topic.getMetadata()
+                            .setNamespace(namespace.get().getMetadata().getName());
+                    topic.setStatus(Topic.TopicStatus.ofSuccess("Imported from cluster"));
+                    return topic;
+                })
+                .filter(Objects::nonNull)
+                // Keep Kafka Streams topics
+                .filter(topic -> topic.getMetadata().getName().endsWith("-changelog")
```
ThomasCAI-mlv: Considering the number of topics in the broker, it might be better to filter the changelog & repartition topics before the map. What do you think?

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -177,10 +186,65 @@ public void synchronizeTopics() {
...
+                    Optional<Namespace> namespace = namespaceService.findByTopicName(
+                            namespaces, topic.getMetadata().getName());
+                    if (namespace.isEmpty()) {
+                        log.debug(
+                                "No namespace found for topic {}. Skipping import.",
+                                topic.getMetadata().getName());
+                        return null;
+                    }
+
+                    topic.getMetadata().setCreationTimestamp(Date.from(Instant.now()));
```
ThomasCAI-mlv: All these operations can be put in `importKafkaStreamsTopics` except for `topic.getMetadata().setNamespace()`, it is more optimized I think

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -177,10 +186,65 @@ public void synchronizeTopics() {
...
+                        .equals(topic.getMetadata().getName())))
+                .toList();
+    }
+
+    /**
+     * Import unsynchronized Kafka Streams topics from broker to Ns4Kafka.
+     *
+     * @param kafkaStreamsBrokerTopics The list of Kafka Streams topics to import
+     */
+    public void importKafkaStreamsTopics(List<Topic> kafkaStreamsBrokerTopics) {
```
ThomasCAI-mlv: This method can be used for non-KStream topics, so how about naming it `importTopics()` and even putting it in topic service?

By the way, I'm thinking about refactoring with the `/_/import` endpoint from the topic controller, it can be done in another PR

loicgreffier: Fixed

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
loicgreffier: Fixed

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
loicgreffier: Done in this PR

ThomasCAI-mlv (approved):

PR #578, opened 2025-06-12
Fix internal stream topics overlap deletion

loicgreffier:


src/main/java/com/michelin/ns4kafka/service/StreamService.java
```diff
@@ -135,17 +136,32 @@ public void delete(Namespace namespace, KafkaStream stream)
                 AccessControlEntryAsyncExecutor.class,
                 Qualifiers.byName(stream.getMetadata().getCluster()));
         accessControlEntryAsyncExecutor.deleteKafkaStreams(namespace, stream);
-        var streamTopicList =
+
+        List<KafkaStream> otherKafkaStreams = findAllForNamespace(namespace).stream()
+                .filter(kafkaStream -> !kafkaStream
+                        .getMetadata()
+                        .getName()
+                        .equals(stream.getMetadata().getName()))
```
ThomasCAI-mlv: How about only getting the kafkaStreams with overlapping name, so that the list is minimal and we have less iteration in the filter l156?
I suggest this condition `kafkaStream.getMetadata().getName().startWith(stream.getMetadata().getName() + "-")` (non negative)

src/main/java/com/michelin/ns4kafka/service/StreamService.java
```diff
@@ -135,17 +136,32 @@ public void delete(Namespace namespace, KafkaStream stream)
...
+        List<Topic> kafkaStreamsTopics =
                 topicService
                         .findByWildcardName(
                                 namespace, stream.getMetadata().getName().concat("-*"))
                         .stream()
                         .filter(topic -> topic.getMetadata().getName().endsWith("-repartition")
                                 || topic.getMetadata().getName().endsWith("-changelog"))
+                        // Exclude topics covered by other Kafka Streams
+                        // (E.g., When deleting "abc.appId", avoid deleting "abc.appId-1234")
+                        .filter(topic -> otherKafkaStreams.stream().noneMatch(kafkaStream -> topic.getMetadata()
```
ThomasCAI-mlv: Unit test for the case when this list is empty: no overlapping name in other kafkastream, so all found topics with `-changelog`and `-repartition` can be deleted safely

src/main/java/com/michelin/ns4kafka/service/StreamService.java
loicgreffier: Fixed

src/main/java/com/michelin/ns4kafka/service/StreamService.java
loicgreffier: Done

ThomasCAI-mlv (approved):
LGTM

PR #646, opened 2025-10-07
Add the possibility to enable transactions for namespaces

ThomasCAI-mlv:


src/main/java/com/michelin/ns4kafka/service/executor/AccessControlEntryAsyncExecutor.java
```diff
@@ -189,7 +180,8 @@ private List<AclBinding> collectBrokerAcls(boolean managedUsersOnly)
             throws ExecutionException, InterruptedException, TimeoutException {
         List<ResourceType> validResourceTypes = List.of(
                 org.apache.kafka.common.resource.ResourceType.TOPIC,
-                org.apache.kafka.common.resource.ResourceType.GROUP);
+                org.apache.kafka.common.resource.ResourceType.GROUP,
+                org.apache.kafka.common.resource.ResourceType.TRANSACTIONAL_ID);
```
loicgreffier, line 184: With transaction ACLs now retrieved from the broker and removed from `buildAclBindingsFromKafkaStream`, all Kafka Streams transaction ACLs will be dropped. Even if they’re covered by the new ACLs, there will be a drop-and-create period that's not acceptable

src/main/java/com/michelin/ns4kafka/service/executor/AccessControlEntryAsyncExecutor.java
```diff
@@ -311,34 +321,26 @@ private List<AclBinding> buildAclBindingsFromKafkaStream(KafkaStream stream, Str
                                 stream.getMetadata().getName(),
                                 PatternType.PREFIXED),
                         new org.apache.kafka.common.acl.AccessControlEntry(
-                                USER_PRINCIPAL + kafkaUser, "*", AclOperation.DELETE, AclPermissionType.ALLOW)),
-                // WRITE on TransactionalId
-                new AclBinding(
```
loicgreffier, line 316: See my previous comment, we might need to keep this one for backward compatibility.

Otherwise:
- All Kafka Streams ACLs will be dropped (if ACL drop is enabled)
- Deleting a KafkaStreams resource won't remove its associated transaction ACL anymore

src/main/java/com/michelin/ns4kafka/service/executor/AccessControlEntryAsyncExecutor.java
ThomasCAI-mlv: The new ACLs will be created before the Kafka Streams ACL are dropped:
The `createAcls(toCreate);` sync method is before the `deleteAcls(toDelete);` method.

It might be still possible that the new ACLs are not taken into account immediatly after they are created?

src/main/java/com/michelin/ns4kafka/service/executor/AccessControlEntryAsyncExecutor.java
ThomasCAI-mlv: For backward compatibility issues, "enabling transactions" is the default behaviour so upgrading ns4kafka to the newest version without specifically disabling transactions will generate all the transactional ID ACLs required for the Kafka Streams.

I agree that if the configuration `drop-unsync-acls` is set at "false", the old Kafka Streams related transactional ID ACLs won't be removed, even after deleting the associated Kafka Stream resource. Do you think it is an issue?

PR #696, opened 2026-01-05
Disable stream topic sync by default and use dedicated HTTP client for Connect health checks

loicgreffier:
…for Connect health checks

### DIFF
--- a/src/test/resources/application-test.yml
+++ b/src/test/resources/application-test.yml
@@ -24,6 +24,7 @@ ns4kafka:
       manage-connectors: true
       manage-topics: true
       manage-users: true
+      sync-kstream-topics: true
       provider: "SELF_MANAGED"
       timeout:
         acl:


src/main/java/com/michelin/ns4kafka/service/client/connect/KafkaConnectClient.java
```diff
@@ -66,6 +65,10 @@ public class KafkaConnectClient {
     @Client(id = "kafka-connect")
     private HttpClient httpClient;
 
+    @Inject
```
ThomasCAI-mlv, line 68: Is this `@Inject` necessary?

loicgreffier: @ThomasCAI-mlv Yep, mandatory for now to inject the related client.

Maybe one day we should consider using constructor injection throughout the whole app instead

Issue #684, opened 2025-12-13
Random connection closed before response was received in connector executor


loicgreffier:
Reproduced on localhost using the demo docker-compose.

During connector synchronization, when collecting connectors from a Kafka Connect instance, Ns4Kafka randomly encounters the following error:

```java
22:59:26.563 [connect-nioEventLoopGroup-4-3] ERROR c.m.n.s.e.ConnectorAsyncExecutor - Error during connectors synchronization for Kafka cluster local and Kafka Connect local: Client 'kafka-connect': Connection closed before response was received.
```

The stack trace is:

```java
io.micronaut.http.client.exceptions.ResponseClosedException: Client 'kafka-connect': Connection closed before response was received
	at io.micronaut.http.client.netty.Http1ResponseHandler$ReaderState.channelInactive(Http1ResponseHandler.java:114)
	at io.micronaut.http.client.netty.Http1ResponseHandler.channelInactive(Http1ResponseHandler.java:85)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:250)
	at io.netty.channel.ChannelInboundHandlerAdapter.channelInactive(ChannelInboundHandlerAdapter.java:81)
	at io.netty.handler.codec.http.HttpContentDecoder.channelInactive(HttpContentDecoder.java:241)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:250)
	at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.fireChannelInactive(CombinedChannelDuplexHandler.java:416)
	at io.netty.handler.codec.ByteToMessageDecoder.channelInputClosed(ByteToMessageDecoder.java:412)
```

This error is triggered at the following line: https://github.com/michelin/ns4kafka/blob/29c556c593d7ba9853f7bebdb43f58e537977847/src/main/java/com/michelin/ns4kafka/service/executor/ConnectorAsyncExecutor.java#L166

When this happens, Kafka Connect sends a FIN request instead of responding normally:

```console
139851	1552.898391	127.0.0.1	127.0.0.1	HTTP	187	GET /connectors?expand=info&expand=status HTTP/1.1 
139852	1552.898430	127.0.0.1	127.0.0.1	TCP	44	8083 → 52884 [ACK] Seq=732 Ack=787 Win=64512 Len=0
[...]

PR #706, opened 2026-01-13
Fix connect synchro thread accumulation in memory

ThomasCAI-mlv:
**Context**
Due to connect synchronisation intervals= being too short, it seemed that the Connect synchro threads are accumulating in the application memory, leading to memory leak (java Heap space)

Locally with Intellij Profiler, the retained bytes for `io.netty.util.concurrent.FastThreadLocalThread` increased from 7.78 MB to 16.98 MB after 30 minutes of runtime.
image

**Implementation proposition**
This PR rolls back the interval duration change for the connector executor, so there is more time for the garbage collector to close the thread before starting new threads.

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/service/executor/KafkaAsyncExecutorScheduler.java
+++ b/src/main/java/com/michelin/ns4kafka/service/executor/KafkaAsyncExecutorScheduler.java
@@ -73,7 +73,7 @@ public class KafkaAsyncExecutorScheduler {
 
     /** Schedule connector synchronization. */
     public void scheduleConnectorSynchronization() {
-        Flux.interval(Duration.ofSeconds(12), Duration.ofSeconds(20))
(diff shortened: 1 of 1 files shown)


src/main/java/com/michelin/ns4kafka/service/executor/KafkaAsyncExecutorScheduler.java
```diff
@@ -73,7 +73,7 @@ public void schedule() {
 
     /** Schedule connector synchronization. */
     public void scheduleConnectorSynchronization() {
-        Flux.interval(Duration.ofSeconds(12), Duration.ofSeconds(20))
+        Flux.interval(Duration.ofSeconds(12), Duration.ofSeconds(30))
```
loicgreffier, line 76: You might try solution 1) then: https://github.com/michelin/ns4kafka/issues/684#issuecomment-3651770874

Otherwise, we’re going to fall back to this previous issue

PR #720, opened 2026-02-09
Add endpoint to delete consumer groups

AfrinKhan02:
### Context

This feature is useful for cleaning up groups that are no longer in use:

To maintain the project's ownership model, the deletion must be restricted:
1.  The namespace must own the consumer group.
2.  The consumer group must be in a safe state to be deleted (`EMPTY`).

### Proposed solution

A new endpoint `DELETE /api/namespaces/{namespace}/consumer-groups/{consumerGroup}` is added.
When called:
1.  It verifies that the namespace is the owner of the consumer group (using existing ACL checks).
2.  It verifies the current state of the consumer group.
    *   It **must** be `EMPTY` (no active members).
    *   If the group is `STABLE` (active) or `DEAD`, the operation is rejected with a `ResourceValidationException`.
3.  If valid, it triggers the deletion via the Kafka Admin Client.
4.  A `dryrun` query parameter is supported to simulate the validation without performing the deletion.

### Implementation

*   **`ConsumerGroupAsyncExecutor`**: Added `deleteConsumerGroups` method to wrap the underlying Kafka Admin Client call.
*   **`ConsumerGroupService`**: Added `deleteConsumerGroup` method to orchestrate the deletion logic.
*   **`ConsumerGroupController`**: Added the `DELETE` endpoint with:
    *   Ownership validation (`isNamespaceOwnerOfConsumerGroup`).
    *   State validation (rejects if not `EMPTY`).
    *   `dryrun` support.
*   **`Kind`**: Added `CONSUMER_GROUP` enum to support the new validation exception type.
*   **`FormatErrorUtils`**: Added `invalidConsumerGroupDeleteOperation` to standardize error messages.

### Remark

The endpoint strictly adheres to the rule that only `EMPTY` groups can be deleted to prevent accidental disruption of active applications.

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/util/FormatErrorUtils.java
+++ b/src/main/java/com/michelin/ns4kafka/util/FormatErrorUtils.java
@@ -162,7 +162,7 @@ public class FormatErrorUtils {
      */
     public static String invalidConnectClusterDeleteOperation(String connectCluster, List<Connector> connectors) {
         return INVALID_OPERATION.formatted(
-                "delete",
+                OPERATION_DELETE,
                 String.format(
                         "The Kafka Connect \"%s\" has %s deployed connector(s): %s. "
                                 + "Please remove the associated connector(s) before deleting it",


src/main/java/com/michelin/ns4kafka/controller/ConsumerGroupController.java
```diff
@@ -138,4 +142,41 @@ public List<ConsumerGroupResetOffsetsResponse> resetOffsets(
...
+    @Delete("/{consumerGroup}{?dryrun}")
+    public HttpResponse<Void> deleteConsumerGroup(
+            String namespace, String consumerGroup, @QueryValue(defaultValue = "false") boolean dryrun)
+            throws ExecutionException, InterruptedException {
+        if (!consumerGroupService.isNamespaceOwnerOfConsumerGroup(namespace, consumerGroup)) {
+            throw new ResourceValidationException(CONSUMER_GROUP, consumerGroup, invalidOwner("group", consumerGroup));
+        }
+
+        Namespace ns = getNamespace(namespace);
+        GroupState currentState = consumerGroupService.getConsumerGroupStatus(ns, consumerGroup);
```
ThomasCAI-mlv, line 184: When the consumer group does not exist, `consumerGroupService.getConsumerGroupStatus` returns `GroupState.Empty` so the endpoint runs `consumerGroupService.deleteConsumerGroup` and throw the exception "500 The group id does not exist."

src/main/java/com/michelin/ns4kafka/controller/ConsumerGroupController.java
```diff
@@ -138,4 +142,41 @@ public List<ConsumerGroupResetOffsetsResponse> resetOffsets(
...
+
+        if (currentState != GroupState.EMPTY) {
```
ThomasCAI-mlv, line 190: It might be better to split the cases when: 
- The consumer group does not exist
- The consumer is currently active/rebalancing
What do you think?

src/main/java/com/michelin/ns4kafka/controller/ConsumerGroupController.java
```diff
@@ -156,4 +161,47 @@ public List<ConsumerGroupResetOffsetsResponse> resetOffsets(
...
+            String namespace, String consumerGroup, @QueryValue(defaultValue = "false") boolean dryrun)
+            throws ExecutionException, InterruptedException {
+        if (!consumerGroupService.isNamespaceOwnerOfConsumerGroup(namespace, consumerGroup)) {
+            throw new ResourceValidationException(CONSUMER_GROUP, consumerGroup, invalidOwner("group", consumerGroup));
+        }
+
+        Namespace ns = getNamespace(namespace);
+
+        if (!consumerGroupService.isConsumerGroupExisting(ns, consumerGroup)) {
+            throw new ResourceValidationException(
```
ThomasCAI-mlv: `ResourceValidationException` will be `422 Unprocessable Entity` status. In case of non-existing consumer group, I think it could be better to return `404 not found` with `HttpResponse.notFound()`

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -128,6 +128,26 @@ public List<String> validateResetOffsets(ConsumerGroupResetOffsets consumerGroup
...
 
+    /**
+     * Check if a consumer group exists.
+     *
+     * @param namespace The namespace
+     * @param groupId The consumer group
+     * @return true if it exists, false otherwise
+     * @throws InterruptedException Any interrupted exception during consumer groups description
+     */
+    public boolean isConsumerGroupExisting(Namespace namespace, String groupId) throws InterruptedException {
```
ThomasCAI-mlv: It seems like duplicate code with the method `getConsumerGroupStatus(Namespace namespace, String groupId)`

For reminder:
- In Kafka 4, `consumerGroupAsyncExecutor.describeConsumerGroups` throws an ExecutionException when the consumer group does not exist 
- In Kafka 3, `consumerGroupAsyncExecutor.describeConsumerGroups` returns `GroupState.DEAD`

One idea in order not to duplicate could be to have `getConsumerGroupStatus` return `GroupState.DEAD` instead of `GroupState.EMPTY` in the catch, so that it does not impact the "reset-offset" endpoint behaviour, and that it can be used in the "delete consumer group" logic.

What do you think?

src/test/java/com/michelin/ns4kafka/controller/ConsumerGroupControllerTest.java
```diff
@@ -284,4 +287,106 @@ void shouldNotResetConsumerGroupWhenItIsStable() throws InterruptedException {
...
+        when(consumerGroupService.getConsumerGroupStatus(ns, "groupID")).thenReturn(GroupState.EMPTY);
+
+        HttpResponse<Void> response = consumerGroupController.deleteConsumerGroup("test", "groupID", false);
+
+        assertEquals(HttpStatus.NO_CONTENT, response.getStatus());
+        verify(consumerGroupService).deleteConsumerGroup(ns, "groupID");
+    }
+
+    @Test
+    void shouldDeleteConsumerGroupDryRun() throws InterruptedException, ExecutionException {
```
ThomasCAI-mlv: Could be renamed `shouldNotDeleteConsumerGroupWhenDryRun`

src/test/java/com/michelin/ns4kafka/controller/ConsumerGroupControllerTest.java
```diff
@@ -284,4 +287,94 @@ void shouldNotResetConsumerGroupWhenItIsStable() throws InterruptedException {
...
+        assertEquals("groupID", result.getName());
+        assertEquals(
+                "Invalid \"delete\" operation: consumer group \"groupID\" can only be deleted if it is "
+                        + GroupState.EMPTY.toString().toLowerCase() + " but the current state is "
+                        + GroupState.STABLE.toString().toLowerCase() + ".",
+                result.getValidationErrors().getFirst());
+    }
+
+    @Test
+    void shouldReturnNotFoundWhenConsumerGroupIsDead() throws InterruptedException, ExecutionException {
```
ThomasCAI-mlv: I think you can keep the same pattern for the test name ->
"shouldActionWhen"
`shouldNotDeleteConsumerGroupWhenDead`

src/main/java/com/michelin/ns4kafka/util/FormatErrorUtils.java
```diff
@@ -324,6 +324,23 @@ public static String invalidConsumerGroupOperation(String consumerGroup, String
...
+     *
+     * @param consumerGroup the consumer group
+     * @param targetState the target state
+     * @param currentState the current state
+     * @return the error message
+     */
+    public static String invalidConsumerGroupDeleteOperation(
+            String consumerGroup, String targetState, String currentState) {
+        return INVALID_OPERATION.formatted(
+                "delete",
```
ThomasCAI-mlv: As SonarCloud raises the issue, can you replace the "delete" by a constant variable

PR #726, opened 2026-02-16
Optimize ACLs synchronization

ThomasCAI-mlv:
**Implementation proposal**
This PR attempts to optimize the ACLs executor:

- Uses `Set` (`HashSet` from Stream API `.collect(Collectors.toSet()`) instead of `List` for the full list of ACLs, to have `O(1)` time complexity for the `.contains()` method used to compute the `toCreate` and `toDelete` ACLs lists
- Uses `Set` for the managed users list for the same reason
- Does the computation of the `toDelete` only if the `drop.unsync.acls` parameter is activated
- Calls the AdminClient to create/delete ACLs only if the `toCreate` and `toDelete` list is not empty
- Splits the ACLs conversion between KafkaStreams and Transactions (for EOS connectors)
- Uses logging guard and Fluent API for trace logs in order not to calculate some ACLs lists when not at the correct log level: the logging guard is used because variable used in lambda must be final
- Uses Stream API `.anyMatch()` operation to find if namespace has any KafkaStream, as it returns the result faster

**Before this PR**
For each ns4kafka GROUP OWNER ACL on resource "group"
- if the associated namespace has TransactionsEnabled or has any KafkaStream, there will be 4 ACLs:
  - WRITE PREFIXED transactional Id `group` 
  - DESCRIBE PREFIXED transactional Id `group` 
  - WRITE PREFIXED transactional Id `connect-group` 
  - DESCRIBE PREFIXED transactional Id `connect-group` 

**After this PR**
For each ns4kafka GROUP OWNER ACL on resource "group"
- if the associated namespace has TransactionsEnabled, there will be 4 ACLs:
  - WRITE PREFIXED transactional Id `group` 
  - DESCRIBE PREFIXED transactional Id `group` 
  - WRITE PREFIXED transactional Id `connect-group` 
  - DESCRIBE PREFIXED transactional Id `connect-group`
- if not and if the namespace has any KafkaStream, only those 2 ACLs will be created:
  - WRITE PREFIXED transactional Id `group` 
  - DESCRIBE PREFIXED transactional Id `group`

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/service/executor/AccessControlEntryAsyncExecutor.java
+++ b/src/main/java/com/michelin/ns4kafka/service/executor/AccessControlEntryAsyncExecutor.java
@@ -227,9 +227,9 @@ public class AccessControlEntryAsyncExecutor {
                                             USER_PRINCIPAL + namespace.getSpec().getKafkaUser()))
                     .collect(Collectors.toSet());
 
-            // Filter out the ACLs to retain only those matching
+            // Remove ACLs of users not managed by Ns4kafka
             userAcls.removeIf(
-                    aclBinding -> managedUsers.contains(aclBinding.entry().principal()));
+                    aclBinding -> !managedUsers.contains(aclBinding.entry().principal()));
         }
 
         // Use logging guard so the ACLs list is only computed when at trace log level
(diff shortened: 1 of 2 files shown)


ThomasCAI-mlv:
The 4 new issues regarding order of imports can be fixed in another PR

PR #732, opened 2026-02-20
Fix initialization progress reporting

loicgreffier:


ThomasCAI-mlv (approved):
Good catch & LGTM! The delay was working well with the "field injection" but not with the constructor injection

PR #740, opened 2026-02-27
Optimize topic import API

ThomasCAI-mlv:
## Implementation proposal
- Initialize and use the "topic OWNER" ACLs list instead of `isNamespaceOwnerOfTopic` for each topic (which iterates over all ACLs to find the namespace ACLs)
- Use `Set` for Ns4kafka topics to have O(1) lookup, instead of `findByName()` for each namespace topic (which iterates on the namespace topics list until the topic is found)
- Regroup all the filters in one method to have a smaller number of topics retrieved from the AdminClient with name parameter filter

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/service/TopicService.java
+++ b/src/main/java/com/michelin/ns4kafka/service/TopicService.java
@@ -42,7 +42,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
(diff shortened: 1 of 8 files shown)


src/main/java/com/michelin/ns4kafka/service/TopicService.java
```diff
@@ -271,39 +271,22 @@ public List<Topic> listUnsynchronizedTopicsByWildcardName(Namespace namespace, S
...
-        List<String> topicNames = listUnsynchronizedTopicNames(namespace);
-
-        // Get topics definitions
-        Collection<Topic> unsynchronizedTopics =
-                topicAsyncExecutor.collectBrokerTopicsFromNames(topicNames).values();
+        List<AccessControlEntry> acls =
+                aclService.findResourceOwnerGrantedToNamespace(namespace, AccessControlEntry.ResourceType.TOPIC);
 
-        return new ArrayList<>(unsynchronizedTopics)
+        Set<String> ns4KafkaTopicNames =
```
loicgreffier: @ThomasCAI-mlv Is this really better to load the whole set of topics once again? I understand it replaces the `.filter(topic -> findByName(namespace, topic).isEmpty())`, but `findByName` is definitely better according to me.

The real issue is that the current `findByName` can be replaced by:

```java
public Optional<Topic> findByName(String cluster, String name) {
    return Optional.ofNullable(getKafkaStore().get(cluster + "/" + name));
}
```

👆 This way we can check if a topic is present in Ns4Kafka or not, without any iteration. BTW we should do the same in Kafka Streams topics import.

See https://github.com/michelin/ns4kafka/pull/744

ThomasCAI-mlv: I definitely agree that the current `findByName` is not efficient for lookup, so I tried a balance between computation time and storage. Your change on `findByName` indeed makes the Set irrelevant

PR #744, opened 2026-03-01
Remove topic name collection from topic import

loicgreffier:


src/main/java/com/michelin/ns4kafka/service/TopicService.java
```diff
@@ -107,13 +106,11 @@ public List<Topic> findByWildcardName(Namespace namespace, String name) {
...
      * @param namespace The namespace
-     * @param topic The topic name
+     * @param topicName The topic name
      * @return An optional topic
      */
-    public Optional<Topic> findByName(Namespace namespace, String topic) {
-        return findAllForNamespace(namespace).stream()
-                .filter(t -> t.getMetadata().getName().equals(topic))
-                .findFirst();
+    public Optional<Topic> findByName(Namespace namespace, String topicName) {
```
loicgreffier, line 112: @ThomasCAI-mlv ACLs are not checked in the function anymore, but I think it’s fine since every calling function basically checks whether the given namespace `isOwnerOf` the given topic.

ThomasCAI-mlv: I agree it is fine, there should be no topic which can be registered under a namespace it doesn't belong to

PR #752, opened 2026-03-06
Add the possibility to force delete connect cluster and connector

AfrinKhan02:
## Context

When a self-deployed Kafka Connect cluster becomes unhealthy (e.g., connection refused, unauthorized), ns4kafka enters an incoherent state:

- **Connect Cluster deletion** is blocked if any connectors remain in the namespace (known by ns4kafka), even if the cluster itself is unreachable.
- **Connector deletion** fails because ns4kafka attempts to call the remote Kafka Connect API to remove the connector, which is unreachable.

This leaves users unable to clean up stale resources from ns4kafka when the underlying Connect infrastructure is down or misconfigured.

## Proposed Solution

A new `force` query parameter (`?force=true`, default `false`) has been added to both Connect Cluster and Connector delete endpoints:

- **Delete Connect Cluster with `force=true`**: Deletes the Connect Cluster resource from ns4kafka **regardless of whether connectors still exist** in the namespace. No cascading deletion of connectors is performed — they remain as independent resources.
- **Delete Connector with `force=true`**: Deletes the Connector resource from the ns4kafka repository **without calling the remote Kafka Connect API**. This allows cleanup even when the Connect cluster is unreachable.

When `force=false` (default), existing behavior is fully preserved.

### `ConnectClusterController`
- Added `@QueryValue(defaultValue = "false") boolean force` parameter to both `delete()` and `bulkDelete()` endpoints.
- The connector-existence validation check is now skipped when `force=true`, allowing the Connect Cluster to be removed independently.

### `ConnectorController`
- Added `@QueryValue(defaultValue = "false") boolean force` parameter to both `delete()` and `bulkDelete()` endpoints.
- The `force` flag is passed through to `ConnectorService.delete()`.

### `ConnectorService`
- Updated `delete(Namespace, Connector, boolean force)` method signature.
- When `force=true`, the remote `kafkaConnectClient.delete()` call is **completely bypassed** — only the local `connectorRepository.delete()` is executed.
- A `WARN`-level log is emitted when force-deleting to provide an audit trail.

### Tests
- All existing tests updated and new test cases created to match the new method signatures (added `boolean force` parameter).

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/service/ConnectorService.java
+++ b/src/main/java/com/michelin/ns4kafka/service/ConnectorService.java
@@ -250,17 +250,18 @@ public class ConnectorService {
                         connector.getSpec().getConnectCluster(),
                         connector.getMetadata().getName())
                 .defaultIfEmpty(HttpResponse.noContent())
-                .doOnError(error -> {
+                .onErrorResume(error -> {
                     if (force) {
-                        connectorRepository.delete(connector);
                         log.atInfo()
                                 .addArgument(connector.getMetadata().getName())
                                 .addArgument(namespace.getMetadata().getName())
                                 .addArgument(connector.getSpec().getConnectCluster())
                                 .addArgument(error.getMessage())
-                                .log("Force deleting Connector [{}] on Kafka [{}] Connect [{}]:"
-                                        + " failed to delete from the connect cluster [{}].");
+                                .log("Success force deleting Connector [{}] on Namespace [{}] from repository."
+                                        + " Failed to delete from Connect cluster [{}]: [{}]");
+                        return Mono.just(HttpResponse.noContent());
                     }
+                    return Mono.error(error);
                 })
                 .map(httpResponse -> {
                     connectorRepository.delete(connector);
--- a/src/test/java/com/michelin/ns4kafka/service/ConnectorServiceTest.java
+++ b/src/test/java/com/michelin/ns4kafka/service/ConnectorServiceTest.java
@@ -1264,7 +1264,8 @@ class ConnectorServiceTest {
         doNothing().when(connectorRepository).delete(connector);
 
         StepVerifier.create(connectorService.delete(ns, connector, true))
-                .verifyError(HttpClientResponseException.class);
+                .consumeNextWith(response -> assertEquals(HttpStatus.NO_CONTENT, response.getStatus()))
+                .verifyComplete();
 
         verify(kafkaConnectClient).delete(ns.getMetadata().getCluster(), "local-name", "ns-connect1");
         verify(connectorRepository).delete(connector);


src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
```diff
@@ -174,13 +174,17 @@ public Mono<HttpResponse<ConnectCluster>> apply(
...
      * @param dryrun Run in dry mode or not
+     * @param force Force deletion of the connect cluster regardless of its connectors
      * @return A HTTP response
-     * @deprecated use {@link #bulkDelete(String, String, boolean)} instead.
+     * @deprecated use {@link #bulkDelete(String, String, boolean, boolean)} instead.
      */
     @Delete("/{connectCluster}{?dryrun}")
     @Deprecated(since = "1.13.0")
     public HttpResponse<Void> delete(
-            String namespace, String connectCluster, @QueryValue(defaultValue = "false") boolean dryrun) {
```
ThomasCAI-mlv, line 183: This method is deprecated, I don't think it needs to be updated

src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
```diff
@@ -226,13 +230,15 @@ public HttpResponse<Void> delete(
      * @param namespace The current namespace
      * @param name The name parameter
      * @param dryrun Run in dry mode or not
+     * @param force Force deletion of the connect cluster regardless of its connectors
```
ThomasCAI-mlv: The description could be "Force delete or not?"

src/main/java/com/michelin/ns4kafka/controller/ConnectorController.java
```diff
@@ -199,13 +199,17 @@ public Mono<HttpResponse<Connector>> apply(
...
+     * @param force Force deletion of the connector regardless of remote deletion success
      * @return A HTTP response
-     * @deprecated use {@link #bulkDelete(String, String, boolean)} instead.
+     * @deprecated use {@link #bulkDelete(String, String, boolean, boolean)} instead.
      */
     @Delete("/{connector}{?dryrun}")
     @Deprecated(since = "1.13.0")
     public Mono<HttpResponse<Void>> delete(
-            String namespace, String connector, @QueryValue(defaultValue = "false") boolean dryrun) {
+            String namespace,
```
ThomasCAI-mlv: I don't think it is necessary to update a deprecated method

src/main/java/com/michelin/ns4kafka/controller/ConnectorController.java
```diff
@@ -235,13 +239,15 @@ public Mono<HttpResponse<Void>> delete(
      * @param namespace The current namespace
      * @param name The name parameter
      * @param dryrun Run in dry mode or not?
+     * @param force Force deletion of the connectors regardless of remote deletion success
```
ThomasCAI-mlv: The description could be "Force delete or not?"

src/main/java/com/michelin/ns4kafka/service/ConnectorService.java
```diff
@@ -243,24 +243,35 @@ public Connector createOrUpdate(Connector connector) {
...
-                            .addArgument(connector.getSpec().getConnectCluster())
-                            .log("Success removing Connector [{}] on Kafka [{}] Connect [{}]");
+        if (force) {
+            log.atWarn()
+                    .addArgument(connector.getMetadata().getName())
+                    .addArgument(namespace.getMetadata().getName())
+                    .addArgument(connector.getSpec().getConnectCluster())
+                    .log(
+                            "Force deleting Connector [{}] on Kafka [{}] Connect [{}] from ns4kafka repository only, bypassing remote API.");
+            deletePublisher = Mono.just(HttpResponse.noContent());
```
ThomasCAI-mlv: Force deleting connectors should delete the connector from the connect clusters if possible, what do you think?

The only difference would be when the connect cluster is unhealthy, the current connector deletion does not delete the connector from ns4kafka, but force would do it

src/test/java/com/michelin/ns4kafka/service/ConnectorServiceTest.java
```diff
@@ -1208,13 +1208,39 @@ void shouldNotDeleteConnectorWhenConnectClusterReturnsError() {
         when(kafkaConnectClient.delete(ns.getMetadata().getCluster(), "local-name", "ns-connect1"))
                 .thenReturn(Mono.error(new HttpClientResponseException("Error", HttpResponse.serverError())));
 
-        StepVerifier.create(connectorService.delete(ns, connector))
+        StepVerifier.create(connectorService.delete(ns, connector, false))
                 .consumeErrorWith(response -> assertEquals(HttpClientResponseException.class, response.getClass()))
                 .verify();
 
         verify(connectorRepository, never()).delete(connector);
     }
```
ThomasCAI-mlv, line 1217: You can include a new test case: `shouldForceDeleteConnectorWhenConnectUnhealthy`

src/main/java/com/michelin/ns4kafka/service/ConnectorService.java
```diff
@@ -243,24 +243,36 @@ public Connector createOrUpdate(Connector connector) {
      * @param namespace The namespace
      * @param connector The connector
      */
-    public Mono<HttpResponse<Void>> delete(Namespace namespace, Connector connector) {
-        return kafkaConnectClient
-                .delete(
-                        namespace.getMetadata().getCluster(),
-                        connector.getSpec().getConnectCluster(),
-                        connector.getMetadata().getName())
-                .defaultIfEmpty(HttpResponse.noContent())
```
ThomasCAI-mlv, line 252: You can keep that same structure, without instanciating `Mono<HttpResponse<Void>>`
Try using `doOnError`

ThomasCAI-mlv: I mean, doing something like this:
Without using a `Mono<HttpResponse<Void>>` variable


kafkaConnectClient
                .delete(
                        namespace.getMetadata().getCluster(),
                        connector.getSpec().getConnectCluster(),
                        connector.getMetadata().getName())
                .defaultIfEmpty(HttpResponse.noContent())
                .doOnError(_ -> { if (force) { connectorRepository.delete(connector);  } }
                .map(httpResponse -> {
                    connectorRepository.delete(connector);
                    log.atInfo()
                            .addArgument(connector.getMetadata().getName())
                            .addArgument(namespace.getMetadata().getName())
                            .addArgument(connector.getSpec().getConnectCluster())
                            .log("Success removing Connector [{}] on Kafka [{}] Connect [{}]");
                    return httpResponse;
                });

src/main/java/com/michelin/ns4kafka/service/ConnectorService.java
```diff
@@ -243,24 +243,37 @@ public Connector createOrUpdate(Connector connector) {
...
-                    log.atInfo()
+        if (force) {
+            deletePublisher = deletePublisher
+                    .doOnError(error -> log.atWarn()
                             .addArgument(connector.getMetadata().getName())
                             .addArgument(namespace.getMetadata().getName())
                             .addArgument(connector.getSpec().getConnectCluster())
-                            .log("Success removing Connector [{}] on Kafka [{}] Connect [{}]");
+                            .addArgument(error.getMessage())
+                            .log("Force deleting Connector [{}] on Kafka [{}] Connect [{}]:"
```
ThomasCAI-mlv: In case of "force delete" when the Connect is not reachable, the connector is not deleted from the Connect, so your log message in this line doesn't seem to be correct

src/main/java/com/michelin/ns4kafka/service/ConnectorService.java
```diff
@@ -243,24 +243,37 @@ public Connector createOrUpdate(Connector connector) {
...
                         connector.getMetadata().getName())
-                .defaultIfEmpty(HttpResponse.noContent())
-                .map(httpResponse -> {
-                    connectorRepository.delete(connector);
+                .defaultIfEmpty(HttpResponse.noContent());
 
-                    log.atInfo()
+        if (force) {
+            deletePublisher = deletePublisher
+                    .doOnError(error -> log.atWarn()
```
ThomasCAI-mlv: Can be INFO log I think

src/main/java/com/michelin/ns4kafka/service/ConnectorService.java
ThomasCAI-mlv: I did not consider the return when not force deleting with unhealthy Connect, so forget what I said before regarding `doOnError`. here are the different cases:
- Force deleting on healthy Connect: delete connector from repository & Connect + return OK status
- Non-force deleting on healthy Connect: delete connector from repository & Connect + return OK status
- Force deleting on unhealthy Connect: delete connector from repository + return OK status
- Non-force deleting on unhealthy Connect: return error message

`onErrorResume` seems better, as you can choose to return the original error:

```
.onErrorResume(error -> {
    if (force) {
        connectorRepository.delete(connector);
        // logging
        return Mono.just(HttpResponse.ok());
    }
    return Mono.error(error);
});
```

You can add it at the end, so that in case of error, the `.defaultIfEmpty` and `.map` are skipped

PR #765, opened 2026-03-16
Add the possibility to delete connectors on cascade when deleting connect clusters

AfrinKhan02:
## Context

The table at the bottom defines the requirement. 

image

> The first row was implemented in https://github.com/michelin/ns4kafka/pull/752. I have proposed a solution for all the other scenarios.

## Proposed Solution

First added a cascade param to the controller, with this `cascade`, 
- with this cascade ns4kafka deletes the related connectors first, then removes the connect cluster 
- If a connector deletion fails due to some reason (timeout or cannot be reached) during `cascade`, the API returns a `Bad Gateway` response and the connect cluster is not deleted, even though some connector deletions may already have succeeded.
- With `cascade` + `force` connect cluster and all it's associated connectors will be forcefully deleted from ns4kafka

## Implementation

In the delete flow I have added the logic for cascade, while doing that I am also using `Mono` & `Flux` for optimizing asynchronous behaviors.

## Tests

Added controller coverage for cascade delete success, cascade failure when Connect is unreachable, and cascade with force. Also added ability to return custom HTTP Exceptions

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/controller/ExceptionHandlerController.java
+++ b/src/main/java/com/michelin/ns4kafka/controller/ExceptionHandlerController.java
@@ -31,6 +31,7 @@ import io.micronaut.http.HttpStatus;
 import io.micronaut.http.annotation.Controller;
 import io.micronaut.http.annotation.Error;
 import io.micronaut.http.exceptions.HttpStatusException;
+import io.micronaut.http.server.exceptions.NotAllowedException;
 import io.micronaut.security.authentication.AuthenticationException;
 import io.micronaut.security.authentication.AuthorizationException;
 import jakarta.validation.ConstraintViolation;
@@ -39,6 +40,7 @@ import jakarta.validation.ElementKind;
 import jakarta.validation.Path;
 import java.util.Iterator;
 import java.util.List;
+import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 
 /** Exception handler controller. */
@@ -109,6 +111,27 @@ public class ExceptionHandlerController {
     }
 
     /**
+     * Handle method not allowed exception.
+     *
+     * @param request the request
+     * @param exception the exception
+     * @return the http response
+     */
+    @Error(global = true)
+    public HttpResponse<Status> error(HttpRequest<?> request, NotAllowedException exception) {
+        var status = Status.builder()
+                .status(FAILED)
+                .message(exception.getMessage())
+                .httpStatus(HttpStatus.METHOD_NOT_ALLOWED)
+                .build();
+
+        return HttpResponse.notAllowed(exception.getAllowedMethods().stream()
+                        .map(io.micronaut.http.HttpMethod::parse)
+                        .collect(Collectors.toSet()))
+                .body(status);
+    }
+
+    /**
      * Handle authentication exception.
      *
      * @param request the request
@@ -138,7 +161,7 @@ public class ExceptionHandlerController {
         if (exception.isForbidden()) {
             var status = Status.builder()
                     .status(FAILED)
-                    .message("Resource forbidden")
+                    .message(exception.getMessage() != null ? exception.getMessage() : HttpStatus.FORBIDDEN.getReason())
                     .httpStatus(HttpStatus.FORBIDDEN)
                     .build();
 
@@ -229,7 +252,7 @@ public class ExceptionHandlerController {
 
         Status status = Status.builder()
                 .status(FAILED)
-                .message("Internal server error")
+                .message(exception.getMessage() != null ? exception.getMessage() : "Internal server error")
                 .httpStatus(HttpStatus.INTERNAL_SERVER_ERROR)
                 .details(StatusDetails.builder()
                         .causes(List.of(exception.getMessage() != null ? exception.getMessage() : exception.toString()))
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -84,6 +84,7 @@ micronaut:
   security:
     authentication: "bearer"
     enabled: true
+    reject-not-found: false
     endpoints:
       introspection:
         enabled: true
--- a/src/test/java/com/michelin/ns4kafka/controller/ExceptionHandlerControllerTest.java
+++ b/src/test/java/com/michelin/ns4kafka/controller/ExceptionHandlerControllerTest.java
@@ -26,10 +26,12 @@ import io.micronaut.http.HttpMethod;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.HttpStatus;
 import io.micronaut.http.exceptions.HttpStatusException;
+import io.micronaut.http.server.exceptions.NotAllowedException;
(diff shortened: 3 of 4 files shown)


src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
```diff
@@ -243,31 +248,58 @@ public HttpResponse<List<ConnectCluster>> bulkDelete(
...
-
-            connectClusterService.delete(cc);
-        });
+        return Flux.fromIterable(connectClusters)
+                .flatMap(cc -> {
+                    List<Connector> connectors = connectorService.findAllByConnectCluster(
+                            ns, cc.getMetadata().getName());
+
+                    Mono<Void> deleteConnectorsMono = Mono.empty();
+                    if (cascade && !connectors.isEmpty()) {
```
ThomasCAI-mlv: If the connectors list is empty, you don't need to check it, as the `Flux.fromIterable(connectors)` will be empty and there will be no execution on the flatMap

src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
```diff
@@ -243,31 +248,58 @@ public HttpResponse<List<ConnectCluster>> bulkDelete(
         connectClusters.forEach(cc -> {
```
ThomasCAI-mlv, line 260: I am wondering if we could store the connectors per connect clusters in a Map, because you are using it multiple times. What do you think?

src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
```diff
@@ -243,31 +248,58 @@ public HttpResponse<List<ConnectCluster>> bulkDelete(
...
         }
 
-        connectClusters.forEach(cc -> {
-            sendEventLog(cc, ApplyStatus.DELETED, cc.getSpec(), null, EMPTY_STRING);
-
-            connectClusterService.delete(cc);
-        });
+        return Flux.fromIterable(connectClusters)
+                .flatMap(cc -> {
+                    List<Connector> connectors = connectorService.findAllByConnectCluster(
```
ThomasCAI-mlv: If you stored the connectors, this line can be replaced

src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
```diff
@@ -243,31 +248,58 @@ public HttpResponse<List<ConnectCluster>> bulkDelete(
...
+                                                                .formatted(
+                                                                        cc.getMetadata()
+                                                                                .getName(),
+                                                                        error.getMessage())
+                                                        + "Please use cascade and force option to bypass the error and remove from Ns4kafka"))
+                                .then();
+                    }
 
-        return HttpResponse.ok(connectClusters);
+                    return deleteConnectorsMono.then(Mono.fromRunnable(() -> {
```
ThomasCAI-mlv: Is `doOnComplete` achieving what you expect?

src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
ThomasCAI-mlv: To avoid the usage of the variable `deleteConnectorsMono`, you can do something like:

```
return (cascade ? Flux.fromIterable(connectors)
                                .flatMap(connectorDeletelogic) : Flux.empty())
.doOnComplete(sendEventLogLogic+CCDeleteLogic)
.then(yourMono)

```

AfrinKhan02:
@ThomasCAI-mlv changes have been made on the basis of comments. Please review again.

src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
```diff
@@ -227,47 +232,79 @@ public HttpResponse<Void> delete(
...
+                                                    error -> new HttpStatusException(
+                                                            HttpStatus.BAD_GATEWAY,
+                                                            "Failed to delete connectors from Connect cluster [%s]: %s. "
+                                                                            .formatted(
+                                                                                    cc.getMetadata()
+                                                                                            .getName(),
+                                                                                    error.getMessage())
+                                                                    + "Please use cascade and force option to bypass the error and remove from Ns4kafka"))
+                                    : Flux.empty())
+                            .then(Mono.fromRunnable(() -> {
```
ThomasCAI-mlv: Is there a specific reason to use `.then(Mono.fromRunnable(myMethod))` instead of `.doOnComplete(myMethod)`?

src/test/java/com/michelin/ns4kafka/controller/connect/ConnectClusterControllerTest.java
```diff
@@ -448,13 +454,130 @@ void shouldBulkForceDeleteConnectClustersWithConnectors() {
...
+        StepVerifier.create(connectClusterController.bulkDelete("test", "connect-cluster", false, false, true))
+                .consumeNextWith(response -> assertEquals(HttpStatus.OK, response.getStatus()))
+                .verifyComplete();
+
+        verify(connectorService).delete(ns, connector, false);
+        verify(connectClusterService).delete(connectCluster);
+    }
+
+    @Test
+    void shouldCascadeDeleteFailWhenConnectUnreachable() {
```
ThomasCAI-mlv: Prefer using `shouldNotVerb` like `shouldNotCascadeDeleteConnectWhenDeletingConnectorFails` or `shouldNotCascadeDeleteConnectWhenConnectorDeletionFails`

src/test/java/com/michelin/ns4kafka/controller/connect/ConnectClusterControllerTest.java
```diff
@@ -448,13 +454,130 @@ void shouldBulkForceDeleteConnectClustersWithConnectors() {
...
+                            "Failed to delete connectors from Connect cluster [connect-cluster]: Host is DOWN. Please use cascade and force option to bypass the error and remove from Ns4kafka",
+                            error.getMessage());
+                })
+                .verify();
+
+        verify(connectClusterService, never()).delete(any());
+    }
+
+    @Test
+    void shouldCascadeForceDeleteWhenConnectUnreachable() {
```
ThomasCAI-mlv: In fact, the connectClusterController is transferring the "force responsibility" to the connectorService, so the test is just `shouldCascadeForceDeleteConnectCluster`

ThomasCAI-mlv (commented):
Hello @AfrinKhan02 , here are the last few comments I have

src/main/java/com/michelin/ns4kafka/controller/connect/ConnectClusterController.java
ThomasCAI-mlv: You can also change this javadoc

AfrinKhan02:
@ThomasCAI-mlv I've addressed all the pr comments , Please review .

PR #775, opened 2026-03-24
Add endpoint to list consumer groups of a namespace

AfrinKhan02:
## Context

This PR implements the "List Consumer Groups" feature for the ns4kafka project. The goal is to allow users to monitor Kafka applications by listing all consumer groups owned by a namespace, including their description and topic offsets.

## Proposed solution

To solve this, I introduced a new API endpoint that lists all consumer groups for a given namespace. The endpoint provides detailed information about each group, including its state and committed offsets, making it much easier to monitor and troubleshoot consumer activity.

## Implementation

- Added a new `ConsumerGroup` model to represent the API response,
- Updated the `ConsumerGroupAsyncExecutor` to provide a method for listing all group IDs, without filtering by group type.
- Implemented `findAllForNamespace` in the service layer, which filters groups by ACL ownership and fetches their descriptions and offsets.
- Exposed a new `GET /consumer-groups` endpoint in `ConsumerGroupController`.
- Registered the new resource in `ApiResourcesController`.
- Refactored the ACL ownership check to reuse the existing `isNamespaceOwnerOfConsumerGroup` method, avoiding duplicated logic.

## Tests

To make sure everything works as expected, I wrote and updated unit tests for both the controller and service. The tests cover normal scenarios, edge cases (like no groups or no ownership), and error handling.

### DIFF
--- a/src/test/java/com/michelin/ns4kafka/service/ConsumerGroupServiceTest.java
+++ b/src/test/java/com/michelin/ns4kafka/service/ConsumerGroupServiceTest.java
@@ -63,7 +63,7 @@ class ConsumerGroupServiceTest {
     ConsumerGroupService consumerGroupService;
 
     @Test
-    void shouldFindConsumerGroupsOwnedByNamespaceWhenListingAll() throws InterruptedException, ExecutionException {
+    void shouldListConsumerGroupsOwnedByNamespace() throws InterruptedException, ExecutionException {
         Namespace namespace = Namespace.builder()
                 .metadata(Resource.Metadata.builder()
                         .name("namespace")
@@ -118,7 +118,7 @@ class ConsumerGroupServiceTest {
     }
 
     @Test
-    void shouldPopulateOffsetsWhenOnlyOneConsumerGroupMatches() throws InterruptedException, ExecutionException {
+    void shouldListConsumerGroupsOwnedByNamespaceWithOffsets() throws InterruptedException, ExecutionException {
         Namespace namespace = Namespace.builder()
                 .metadata(Resource.Metadata.builder()
                         .name("namespace")
@@ -173,7 +173,7 @@ class ConsumerGroupServiceTest {
     }
 
     @Test
-    void shouldFindConsumerGroupsByWildcardName() throws InterruptedException, ExecutionException {
+    void shouldListConsumerGroupsOwnedByNamespaceWithWildcardName() throws InterruptedException, ExecutionException {
         Namespace namespace = Namespace.builder()
                 .metadata(Resource.Metadata.builder()
                         .name("namespace")
@@ -216,7 +216,7 @@ class ConsumerGroupServiceTest {
     }
 
     @Test
-    void shouldReturnEmptyConsumerGroupListWhenNamespaceOwnsNone() throws InterruptedException, ExecutionException {
+    void shouldListConsumerGroupsWhenEmpty() throws InterruptedException, ExecutionException {
         Namespace namespace = Namespace.builder()
                 .metadata(Resource.Metadata.builder()
                         .name("namespace")


src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -63,15 +66,63 @@ public ConsumerGroupService(ApplicationContext applicationContext, AclService ac
...
+     */
+    public List<ConsumerGroup> findAllForNamespace(Namespace namespace)
+            throws ExecutionException, InterruptedException {
+        ConsumerGroupAsyncExecutor consumerGroupAsyncExecutor = applicationContext.getBean(
+                ConsumerGroupAsyncExecutor.class,
+                Qualifiers.byName(namespace.getMetadata().getCluster()));
+
+        List<String> allGroupIds = consumerGroupAsyncExecutor.listConsumerGroupIds();
+
+        List<String> consumerGroupIds = allGroupIds.stream()
```
ThomasCAI-mlv: I think you can directly stream and filter on the result of `listConsumerGroupIds()` since it is not used anywhere else

src/test/java/com/michelin/ns4kafka/controller/ConsumerGroupControllerTest.java
```diff
@@ -78,6 +79,28 @@ class ConsumerGroupControllerTest {
     @InjectMocks
     ConsumerGroupController consumerGroupController;
 
+    @Test
+    void shouldListConsumerGroupsForNamespace() throws InterruptedException, ExecutionException {
```
ThomasCAI-mlv: Test name `shouldListConsumerGroup` is enough, as we are in the namespaced controller. What do you think?

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -285,4 +336,32 @@ public void alterConsumerGroupOffsets(
...
+                .metadata(Metadata.builder()
+                        .name(consumerGroupId)
+                        .namespace(namespace.getMetadata().getName())
+                        .cluster(namespace.getMetadata().getCluster())
+                        .build())
+                .status(ConsumerGroup.ConsumerGroupStatus.builder()
+                        .state(description == null ? GroupState.UNKNOWN : description.groupState())
+                        .offsets(committedOffsets.entrySet().stream()
+                                .sorted(Map.Entry.<TopicPartition, Long>comparingByKey(
+                                                java.util.Comparator.comparing(TopicPartition::topic)
```
ThomasCAI-mlv: Maybe you can import `comparing`

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -285,4 +336,32 @@ public void alterConsumerGroupOffsets(
...
+                                                        .thenComparingInt(TopicPartition::partition))
+                                        .thenComparing(Map.Entry.comparingByValue()))
```
ThomasCAI-mlv: Is it needed to compare by value (offset) at the end? there are only 3 fields

If the initial TopicPartition variable is too loaded / we can `map` first before doing the sorting, like

```
.map(....)
.sorted(comparing(ConsumerGroup.ConsumerGroupOffset::getTopic))
```

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -285,4 +336,32 @@ public void alterConsumerGroupOffsets(
...
+                        .name(consumerGroupId)
+                        .namespace(namespace.getMetadata().getName())
+                        .cluster(namespace.getMetadata().getCluster())
+                        .build())
+                .status(ConsumerGroup.ConsumerGroupStatus.builder()
+                        .state(description == null ? GroupState.UNKNOWN : description.groupState())
+                        .offsets(committedOffsets.entrySet().stream()
+                                .sorted(Map.Entry.<TopicPartition, Long>comparingByKey(
+                                                java.util.Comparator.comparing(TopicPartition::topic)
+                                                        .thenComparingInt(TopicPartition::partition))
```
ThomasCAI-mlv: Sorting by topic name can be useful if the consumer group has many topics, which is unusual I think
And there is no real reason to have an ordering for partition, so we can have it or not. Personally I would just sort by topic name, but you can keep the partition if you want

src/main/java/com/michelin/ns4kafka/controller/ConsumerGroupController.java
```diff
@@ -74,6 +76,19 @@ public ConsumerGroupController(
...
+    /**
+     * List consumer groups owned by the namespace.
+     *
+     * @param namespace The namespace
+     * @return The list of consumer groups
+     * @throws ExecutionException Any execution exception
+     * @throws InterruptedException Any interrupted exception
+     */
+    @Get
+    public List<ConsumerGroup> list(String namespace) throws ExecutionException, InterruptedException {
```
ThomasCAI-mlv: If "consumer group" is part of API resources, it might be better to follow other API resources endpoints: add a `name` query parameter which filters using wildcard or exact match

src/main/java/com/michelin/ns4kafka/controller/ConsumerGroupController.java
```diff
@@ -74,6 +76,19 @@ public ConsumerGroupController(
...
+        return consumerGroupService.findAllForNamespace(getNamespace(namespace));
```
ThomasCAI-mlv: Since there are some external calls with the adminClient (to retrieve offsets per TopicPartition), you might consider doing like the `list` in the SchemaController: 
- If there are exactly 1 consumer group, the offsets per TopicPartition is queried and are returned
- if there are more than 1, only the consumer groups name are returned

This trick avoids too many blocking calls which can hinder Ns4kafka performance (we already had this issue for schemas)

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -63,15 +66,63 @@ public ConsumerGroupService(ApplicationContext applicationContext, AclService ac
...
+
+        if (consumerGroupIds.isEmpty()) {
+            return List.of();
+        }
+
+        Map<String, ConsumerGroupDescription> descriptions =
+                consumerGroupAsyncExecutor.describeConsumerGroups(consumerGroupIds);
+
+        List<ConsumerGroup> consumerGroups = new ArrayList<>();
+        for (String groupId : consumerGroupIds) {
```
ThomasCAI-mlv: Could you use Java Stream API instead of the "for" loop?
```
consumerGroupIds.stream()
.map(logicWithErrorHandling)
```

If the consumerGroupIds is empty, the result will be an empty list, without the `getCommittedOffsets()` external call being executed (which avoids `if (consumerGroupIds.isEmpty()) {
            return List.of();
        }`)

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -63,15 +66,63 @@ public ConsumerGroupService(ApplicationContext applicationContext, AclService ac
...
     /**
      * Check if a given namespace is owner of a given group.
      *
      * @param namespace The namespace
      * @param groupId The group
      * @return true if it is, false otherwise
      */
-    public boolean isNamespaceOwnerOfConsumerGroup(String namespace, String groupId) {
-        return aclService.isNamespaceOwnerOfResource(namespace, AccessControlEntry.ResourceType.GROUP, groupId);
+    public boolean isNamespaceOwnerOfConsumerGroup(Namespace namespace, String groupId) {
```
ThomasCAI-mlv: The initial code was not working or is it a refactoring?

If it is just a refactoring, why not keep `isNamespaceOwnerOfResource()` ?

src/main/java/com/michelin/ns4kafka/controller/ApiResourcesController.java
```diff
@@ -109,6 +109,15 @@ public class ApiResourcesController {
             .names(List.of("connect-clusters", "connect-cluster", "cc", "kconnect", "kconnects", "kc"))
             .build();
 
+    /** Consumer group resource definition. */
```
ThomasCAI-mlv: Can you remove the consumer group from the API resource? It is not an API resource because it does not have the structure of the API resource with the following 3 endpoints:
- GET /{API_RESOURCE}(?name}
- POST /{API_RESOURCE}
- DELETE /{API_RESOURCE}(?name}

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -63,6 +69,35 @@ public ConsumerGroupService(ApplicationContext applicationContext, AclService ac
...
 
+    /**
+     * Find all consumer groups owned by a given namespace.
+     *
+     * @param namespace The namespace
+     * @return A list of consumer groups
+     * @throws ExecutionException Any execution exception during consumer groups listing
+     * @throws InterruptedException Any interrupted exception during consumer groups listing
+     */
+    public List<ConsumerGroup> findAllForNamespace(Namespace namespace)
```
ThomasCAI-mlv: Check comment about `findAllForNamespace` & `findByWildcardname`

Depending on the outcome, this method can be removed along with the associated tests

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -285,4 +333,71 @@ public void alterConsumerGroupOffsets(
...
+                                        .currentOffset(entry.getValue())
+                                        .build())
+                                .sorted(comparing(ConsumerGroup.ConsumerGroupOffset::getTopic)
+                                        .thenComparingInt(ConsumerGroup.ConsumerGroupOffset::getPartition))
+                                .toList())
+                        .build())
+                .build();
+    }
+
+    private List<ConsumerGroup> findConsumerGroups(Namespace namespace, Predicate<String> groupIdFilter)
```
ThomasCAI-mlv: That method can be part of the `findByWildcardName`
I'm not too fond of predicate if it is only used once

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
```diff
@@ -285,4 +333,71 @@ public void alterConsumerGroupOffsets(
...
+            throws ExecutionException, InterruptedException {
+        ConsumerGroupAsyncExecutor consumerGroupAsyncExecutor = applicationContext.getBean(
+                ConsumerGroupAsyncExecutor.class,
+                Qualifiers.byName(namespace.getMetadata().getCluster()));
+
+        List<String> consumerGroupIds = consumerGroupAsyncExecutor.listConsumerGroupIds().stream()
+                .filter(groupId -> isNamespaceOwnerOfConsumerGroup(namespace, groupId))
```
ThomasCAI-mlv: Actually, you can initialize 

```
List<AccessControlEntry> ownerAcls = 
            aclService.findResourceOwnerGrantedToNamespace(namespace, AccessControlEntry.ResourceType.GROUP);
```
And filter using `aclService.isResourceCoveredByAcls()` instead.

The reason is to avoid retrieving OWNER Group ACLs for every consumer group

For other endpoints like reset-offset and delete, you are only managing one consumer group at a time, so it is fine to use `isNamespaceOwnerOfConsumerGroup`

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
ThomasCAI-mlv: Depending on if you prefer it split or not:

You can have 2 methods:
- `findAllForNamespace` which lists all consumer groups and filters on the namespace, then build a ConsumerGroup list without offset
- `findByWildcardName` which uses the result of `findAllForNamespace` and filters on the `name` parameter, then add offset information if there is exactly 1 element

Or 1 method:
- `findByWildcardName` which lists all consumer groups and filters on the namespace & filters on the `name` parameter, then build a ConsumerGroup list with offset information if there is exactly 1 element

In any case, `describeConsumerGroups` can be done in `findByWildcardName`

src/main/java/com/michelin/ns4kafka/service/ConsumerGroupService.java
ThomasCAI-mlv: Ok I see, `isNamespaceOwnerOfResource` is used at too many places, the refactor would be too costly
This method is not needed anymore, as it is preferred to initialize ownerAcls outside

AfrinKhan02:
@ThomasCAI-mlv I've addressed the PR review comments, Can you please review .

ThomasCAI-mlv (approved):
LGTM

PR #790, opened 2026-04-06
Add `RegexPattern` validator with lenient/strict mode and warning propagation

nil-malh:
### What

Introduces a new `RegexPattern`(#788) validation constraint and wires a complete
**soft-failure (warning) pipeline** from the validator layer all the way to
the HTTP response, so a misconfigured resource value can be flagged to the
caller without blocking the operation.

---

### Motivation

Previous validators only had two outcomes: **pass** or **hard error** (the
resource is rejected with HTTP 422). There was no way to express a softer
"this looks wrong but we'll allow it" signal. Teams wanted a regex-based
constraint that could run in two modes:

| Strict | Behaviour |
|------|-----------|
| `true` | Value must match the regex; otherwise the resource is **rejected**. |
| `false` | Value must match the regex; a mismatch surfaces a **warning** but the resource is **created/updated**. |

---

#### New / modified production code

| File | Change |
|------|--------|
| `ResourceValidator.RegexPattern` | New inner-class validator. Accepts `regex` + `strict` (`true` \| `lenient`). In lenient mode throws `FieldValidationException(msg, soft=true)`; in strict mode throws the same exception with `soft=false`. Short-circuits (no-op) if `regex` is null or empty. |
| `FieldValidationException` | Added `boolean soft` field and a two-arg constructor so callers can mark an exception as a soft/warning failure. |
| `ValidationResult` | New `record` that carries two lists: `errors` (hard failures) and `warnings` (soft failures). Includes factory methods `empty()`, `ofErrors(List)` and a `merge(ValidationResult)` helper. |
| `TopicValidator.validate()` | Return type changed `List<String>` → `ValidationResult`. `FieldValidationException` is now routed: `e.soft == true` → `warnings`, `false` → `errors`. |
| `ConnectValidator.validate()` | Same change as `TopicValidator`, applied to all four constraint maps (global, sink, source, class). Early-exit (empty name) now returns `ValidationResult.ofErrors(...)`. |
| `ConnectorService.validateLocally()` | Return type changed `Mono<List<String>>` → `Mono<ValidationResult>`. Early-exit error paths wrapped in `ValidationResult.ofErrors(...)`. |
| `ResourceController` | Added `X-Ns4kafka-Warnings` header constant and a new overloaded `formatHttpResponse(body, status, warnings)` that sets the header when the warning list is non-empty. |
| `TopicController.apply()` | Splits validator output into `validationErrors` + `validationWarnings`. Only throws `ResourceValidationException` on hard errors. All three success return paths (`UNCHANGED`, dry-run, `CREATED`/`CHANGED`) forward warnings via the new `formatHttpResponse` overload. |
| `ConnectorController.apply()` | Same pattern: uses `localResult.hasErrors()` guard, collects `localResult.warnings()` and passes them to every success response path. |

#### Test updates

| File | Change |
|------|--------|
| `ResourceValidatorTest` | Added `shouldValidateRegexPattern` and `shouldSkipValidationWhenRegexNotSet`. Lenient-mode mismatch tests assert that a `FieldValidationException` **is** thrown and that `e.soft == true`. |
| `TopicValidatorTest` | All `validate()` call sites updated to `ValidationResult`; assertions use `.errors()`. |
| `ConnectValidatorTest` | Same as above. Expected error messages now built from the connector's actual name (avoids brittle string-counting). |
| `ConnectorServiceTest` | `validateLocally()` assertions updated to `.errors().size()` / `.errors().getFirst()`; success assertions use `assertFalse(response.hasErrors())`. |
| `ConnectorControllerTest` | `validateLocally` mocks updated: empty-success → `Mono.just(ValidationResult.empty())`; error case → `Mono.just(new ValidationResult(List.of("…"), List.of()))`. |

---

### How warnings reach the client

```
RegexPattern.ensureValid()  →  FieldValidationException(soft=true)
        ↓
TopicValidator / ConnectValidator.validate()
        catches soft=true  →  ValidationResult.warnings
        ↓
TopicController / ConnectorController.apply()
        no hard errors  →  resource saved
        warnings present  →  formatHttpResponse(…, warnings)
        ↓
HTTP response
  X-Ns4kafka-Result:   created | changed | unchanged
  X-Ns4kafka-Warnings: Invalid value "FOO" for field "bar": value must match regex "…"
```

---

### What has NOT changed

- `validateRemotely()` still returns `Mono<List<String>>` (remote Kafka Connect errors are always hard).
- `ResourceValidationException` still only carries hard errors; it is never thrown for warnings.
- No breaking changes to the existing API response body shape.

### To be done 

- [X] Add suports for X-Ns4kafka-Warning in Kafkactl to display warning when applying ([PR here](https://github.com/michelin/kafkactl/pull/343))

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/controller/generic/ResourceController.java
+++ b/src/main/java/com/michelin/ns4kafka/controller/generic/ResourceController.java
@@ -29,13 +29,14 @@ import io.micronaut.http.HttpResponse;
 import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.security.utils.SecurityService;
 import java.time.Instant;
+import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 
 /** Resource controller. */
 public abstract class ResourceController {
     private static final String STATUS_HEADER = "X-Ns4kafka-Result";
-    static final String WARNING_HEADER = "X-Ns4kafka-Warnings";
+    private static final String WARNING_HEADER = "X-Ns4kafka-Warnings";
     protected final SecurityService securityService;
     protected final ApplicationEventPublisher<AuditLog> applicationEventPublisher;
 
@@ -60,7 +61,7 @@ public abstract class ResourceController {
      * @param <T> The type of the response body
      */
     public <T> HttpResponse<T> formatHttpResponse(T body, ApplyStatus status) {
-        return HttpResponse.ok(body).header(STATUS_HEADER, status.toString());
+        return formatHttpResponse(body, status, Collections.emptyList());
     }
 
     /**
@@ -72,7 +73,7 @@ public abstract class ResourceController {
      *
      * @param body The response body
      * @param status The operation status
-     * @param warnings The list of soft-validation warnings (may be empty)
+     * @param warnings The list of soft-validation warnings (maybe empty)
      * @return The formatted HTTP response
      * @param <T> The type of the response body
      */
--- a/src/main/java/com/michelin/ns4kafka/controller/generic/ResourceController.java
+++ b/src/main/java/com/michelin/ns4kafka/controller/generic/ResourceController.java
@@ -26,13 +26,16 @@ import com.michelin.ns4kafka.security.ResourceBasedSecurityRule;
 import com.michelin.ns4kafka.util.enumation.ApplyStatus;
 import io.micronaut.context.event.ApplicationEventPublisher;
 import io.micronaut.http.HttpResponse;
+import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.security.utils.SecurityService;
 import java.time.Instant;
 import java.util.Date;
+import java.util.List;
 
 /** Resource controller. */
 public abstract class ResourceController {
     private static final String STATUS_HEADER = "X-Ns4kafka-Result";
+    static final String WARNING_HEADER = "X-Ns4kafka-Warnings";
     protected final SecurityService securityService;
     protected final ApplicationEventPublisher<AuditLog> applicationEventPublisher;
 
@@ -58,6 +61,27 @@ public abstract class ResourceController {
      */
     public <T> HttpResponse<T> formatHttpResponse(T body, ApplyStatus status) {
         return HttpResponse.ok(body).header(STATUS_HEADER, status.toString());
+    }
+
+    /**
+     * Format an HTTP response with the operation status and optional validation warnings.
+     *
+     * <p>When {@code warnings} is non-empty each warning message is joined with {@code ", "} and emitted in the
+     * {@value WARNING_HEADER} response header so clients (e.g. Kafkactl) can surface them without blocking the
+     * operation.
+     *
+     * @param body The response body
+     * @param status The operation status
+     * @param warnings The list of soft-validation warnings (may be empty)
+     * @return The formatted HTTP response
+     * @param <T> The type of the response body
+     */
+    public <T> HttpResponse<T> formatHttpResponse(T body, ApplyStatus status, List<String> warnings) {
+        MutableHttpResponse<T> response = HttpResponse.ok(body).header(STATUS_HEADER, status.toString());
+        if (warnings != null && !warnings.isEmpty()) {
+            response.header(WARNING_HEADER, String.join(", ", warnings));
+        }
+        return response;
     }
 
     /**
--- a/src/main/java/com/michelin/ns4kafka/validation/ConnectValidator.java
+++ b/src/main/java/com/michelin/ns4kafka/validation/ConnectValidator.java
@@ -18,7 +18,9 @@
  */
 package com.michelin.ns4kafka.validation;
 
-import static com.michelin.ns4kafka.util.FormatErrorUtils.*;
+import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameEmpty;
+import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameLength;
+import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameSpecChars;
 import static com.michelin.ns4kafka.util.config.ConnectorConfig.CONNECTOR_CLASS;
 
 import com.fasterxml.jackson.annotation.JsonSetter;
--- a/src/main/java/com/michelin/ns4kafka/validation/ConnectValidator.java
+++ b/src/main/java/com/michelin/ns4kafka/validation/ConnectValidator.java
@@ -18,9 +18,7 @@
  */
 package com.michelin.ns4kafka.validation;
 
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameEmpty;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameLength;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameSpecChars;
+import static com.michelin.ns4kafka.util.FormatErrorUtils.*;
 import static com.michelin.ns4kafka.util.config.ConnectorConfig.CONNECTOR_CLASS;
 
 import com.fasterxml.jackson.annotation.JsonSetter;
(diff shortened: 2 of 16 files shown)


src/main/java/com/michelin/ns4kafka/validation/ResourceValidator.java
```diff
@@ -337,6 +332,41 @@ public void ensureValid(String name, Object value) {
...
+    /** Validation logic for a given regex pattern on a string. */
+    @Data
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class RegexPattern implements Validator {
+        public static final List<String> VALID_MODES = List.of("lenient", "strict");
+        private String regex;
+        private String mode;
+
+        public static RegexPattern matches(String regex, String mode) {
```
loicgreffier: For `mode`, do you think we could adopt either:

`strict: true|false` (default: true) or `warning: true|false` (default: false) or `soft: true|false` (default false), to stay consistent with the existing optional: true|false option?

This would also be easier to reuse across other validators if needed (not in the scope of this PR, but still an option...) and to forward into `FieldValidationException`

src/main/java/com/michelin/ns4kafka/controller/generic/ResourceController.java
```diff
@@ -26,13 +26,16 @@
...
+import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.security.utils.SecurityService;
 import java.time.Instant;
 import java.util.Date;
+import java.util.List;
 
 /** Resource controller. */
 public abstract class ResourceController {
     private static final String STATUS_HEADER = "X-Ns4kafka-Result";
+    static final String WARNING_HEADER = "X-Ns4kafka-Warnings";
```
loicgreffier: Make it `private static final`

src/main/java/com/michelin/ns4kafka/controller/generic/ResourceController.java
```diff
@@ -60,6 +63,27 @@ public <T> HttpResponse<T> formatHttpResponse(T body, ApplyStatus status) {
         return HttpResponse.ok(body).header(STATUS_HEADER, status.toString());
```
loicgreffier: Can this now calls `formatHttpResponse(body, status, Collections.emptyList()`?

src/main/java/com/michelin/ns4kafka/validation/FieldValidationException.java
```diff
@@ -33,4 +34,16 @@ public class FieldValidationException extends RuntimeException {
...
+    /**
+     * Constructor.
+     *
+     * @param error The error message
+     * @param soft {@code true} if this exception is informational and should be surfaced to the user as a warning
+     *     rather than a hard failure; {@code false} if it represents a critical validation error.
+     */
+    public FieldValidationException(String error, boolean soft) {
+        super(error);
+        this.soft = soft;
```
loicgreffier, line 47: Align the parameter name depending on what we choose here: https://github.com/michelin/ns4kafka/pull/790/changes#r3043672450

src/main/java/com/michelin/ns4kafka/validation/TopicValidator.java
```diff
@@ -99,14 +99,18 @@ public static TopicValidator makeDefaultOneBroker() {
...
+     *
      * @param topic The topic
-     * @return A list of validation errors
+     * @return A {@link ValidationResult} containing errors and warnings
      * @see <a href=
      *     "https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L36">
      *     GitHub</a>
      */
-    public List<String> validate(Topic topic) {
+    public ValidationResult validate(Topic topic) {
```
loicgreffier: Referring to the initial issue: https://github.com/michelin/ns4kafka/issues/788

> being able to have a validator with a regex on the name of each resources

Right now, validations are applied only to `spec` fields only. So we need the actual logic to validate `metadata.name`.

Based on:

```yml
topicValidator:
  validationConstraints:
    name:
      validation-type: "RegexPattern"
      regex: ...
      soft: true
```

We need to add the validation logic on `metadata.name` over there: https://github.com/michelin/ns4kafka/blob/ac14e4fe4a0f7abacc89df6b5de5c6042b03510b/src/main/java/com/michelin/ns4kafka/validation/TopicValidator.java#L137

Same for the connectors.

loicgreffier (commented):
@nil-malh Here is my feedback

src/main/java/com/michelin/ns4kafka/validation/ResourceValidator.java
nil-malh: Agreed :)

src/main/java/com/michelin/ns4kafka/validation/FieldValidationException.java
nil-malh: I think that having `strict` as the name in the validator makes sense. However I am not sure about here for me a soft exception is an exception that throws but do not interrupt the flow, I am not at ease with the `strict` exception this would means that all exceptions are not strict but they are by default strict as they are 99% of the time explicitly thrown. 

`strict` in the exception would also create an odd mental model : 
- A `FieldValidationException(error, strict=false)` reads like "a validation exception that isn't strict", which is contradictory. 
- A `FieldValidationException(error, soft=true)` reads like "a validation exception, but a soft one", which is exactly what it is.
What do you think ? :)

src/main/java/com/michelin/ns4kafka/validation/TopicValidator.java
nil-malh: whooopsies, nice catch ! I'll add it right now ;)

nil-malh: It's done, I have also taken the liberty of refactoring both Validator classes as they were starting to get repetitive

src/main/java/com/michelin/ns4kafka/validation/ConnectValidator.java
```diff
@@ -18,9 +18,7 @@
  */
 package com.michelin.ns4kafka.validation;
 
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameEmpty;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameLength;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidNameSpecChars;
+import static com.michelin.ns4kafka.util.FormatErrorUtils.*;
```
loicgreffier: Avoid wildcard imports

src/main/java/com/michelin/ns4kafka/validation/FieldValidationException.java
loicgreffier: Let's preserve `soft`

src/main/java/com/michelin/ns4kafka/validation/ResourceValidator.java
```diff
@@ -18,13 +18,7 @@
...
 package com.michelin.ns4kafka.validation;
 
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidFieldValidationAtLeast;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidFieldValidationAtMost;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidFieldValidationContains;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidFieldValidationEmpty;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidFieldValidationNull;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidFieldValidationNumber;
-import static com.michelin.ns4kafka.util.FormatErrorUtils.invalidFieldValidationOneOf;
+import static com.michelin.ns4kafka.util.FormatErrorUtils.*;
```
loicgreffier: Avoid wildcard imports

src/test/java/com/michelin/ns4kafka/model/ResourceValidatorTest.java
```diff
@@ -18,10 +18,7 @@
  */
 package com.michelin.ns4kafka.model;
 
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.*;
```
loicgreffier: Avoid wildcard imports

loicgreffier (commented):
@nil-malh Second round, mainly for * imports.

Unless I missed anything, I did not see the validation for the name in the `TopicValidator`?

nil-malh:
> Unless I missed anything, I did not see the validation for the name in the `TopicValidator`?

I absolutely forgot to add it in topic nice catch

loicgreffier (approved):

PR #791, opened 2026-04-13
Fix infinite synchro topic cleanup policy

### LINKED ISSUE #737: Infinite synchro with "delete, compact" topic cleanup policy
## Describe the bug
When a topic is applied with "delete, compact" cleanup policy, the topic is then infinitely synchronized

## To Reproduce

1. Create a topic putting `delete, compact` in the `spec.configs.cleanup.policy` field
2. Check every 20 sec that the topic is applied successfully

## Expected behavior
The topic should be applied (created/updated) once.

## Environment (please complete the following information):

- Ns4Kafka version: 1.18.3
- Kafka broker type and version: Confluent Cloud, v8

## Steps taken to try and solve the problem
By putting `delete,compact` (without the space after the comma), Ns4kafka correctly applies the topic once. So it might be due to the "cleanup policy" format returned by the adminClient which is "delete,compact". This interfere with the comparison.

## Additional context
`compact,delete` needs to be checked as well.

AfrinKhan02:
# Context

This pull request resolves [#737](https://github.com/michelin/ns4kafka/issues/737), where topics configured with `cleanup.policy: "delete, compact"` could be infinitely synchronized.

The issue happened because Ns4Kafka compared the topic configuration from its side with the one returned by Kafka as plain strings. Depending on the broker, `cleanup.policy` could come back as `delete,compact` or `compact,delete`, so Ns4Kafka kept thinking the topic was different and re-applied it on every synchronization cycle.

# Proposed solution

Normalize `cleanup.policy` values before comparing them during topic synchronization and validation.

The goal is to treat equivalent cleanup-policy representations the same way, including:

- `delete, compact`
- `delete,compact`
- `compact,delete`
- `compact, delete`

This keeps synchronization idempotent while preserving the existing Confluent Cloud validation rule for unsupported transitions from `delete` to `delete + compact`.

# Implementation

I introduced `TopicConfigUtils` to centralize cleanup-policy normalization and comparison logic.

The implementation:

- normalizes `cleanup.policy` by splitting on `,`, trimming spaces, sorting the values, and joining them back before comparison
- uses that normalization in `TopicAsyncExecutor` so equivalent cleanup-policy values no longer trigger repeated config updates
- uses the same normalization in `TopicService` so the Confluent Cloud validation still rejects unsupported `delete` -> `delete + compact` updates, even when the input contains spaces or reversed ordering.

# Tests

I updated the test suite to cover the bug scenario and the normalized comparison behavior.

# Remark

This fix keeps the behavior aligned with the issue expectation from [#737](https://github.com/michelin/ns4kafka/issues/737): a topic configured with `delete, compact` should be applied once, then remain stable on subsequent synchronization cycles.


AfrinKhan02:
@ThomasCAI-mlv I've fixed the bug , can you review.

src/main/java/com/michelin/ns4kafka/service/TopicService.java
```diff
@@ -243,9 +244,8 @@ public List<String> validateTopicUpdate(Namespace namespace, Topic existingTopic
                 .orElse(CLEANUP_POLICY_DELETE);
 
         if (isConfluentCloud
-                && existingCleanUpPolicy.equals(CLEANUP_POLICY_DELETE)
-                && (newCleanUpPolicy.equals(CLEANUP_POLICY_DELETE + "," + CLEANUP_POLICY_COMPACT)
-                        || newCleanUpPolicy.equals(CLEANUP_POLICY_COMPACT + "," + CLEANUP_POLICY_DELETE))) {
+                && TopicConfigUtils.isDeleteCleanupPolicy(existingCleanUpPolicy)
```
ThomasCAI-mlv: It seems to me that `existingCleanUpPolicy.equals(CLEANUP_POLICY_DELETE)` is more direct. I think there is no possible confusion for `DELETE` or `COMPACT` policy, so you can save some computations by not doing `normalizeValue` for these policies

ThomasCAI-mlv (commented):
Hello @AfrinKhan02 this is my review, just a little thing and it can be merged!

AfrinKhan02:
@ThomasCAI-mlv I made the addressed changes, Can you review

ThomasCAI-mlv (approved):

PR #817, opened 2026-05-26
Add the ability to stop connectors 

AfrinKhan02:
## Context

This change adds support for stopping a Kafka Connect connector from Ns4Kafka.

Until now, the connector state change flow supported `RESTART`, `PAUSE`, and `RESUME`, but not `STOP`. As a result, users could not stop a connector through the existing `change-state` API even though stop is a valid Kafka Connect operation.

The goal of this PR is to expose connector stop through the existing connector state change workflow.

## Proposed solution

I extended the existing `ChangeConnectorState` flow to support a new `STOP` action.

The objective was to keep the implementation aligned with the current connector action model, so stop uses the same request shape and dispatching logic as the other connector state transitions.

## Implementation

I added `STOP` to the `ChangeConnectorState.ConnectorAction` enum so it can be accepted through the existing `change-state` request body.

I then wired the `STOP` action in the connector controller so `POST /api/namespaces/{namespace}/connectors/{connector}/change-state` dispatches to the connector service when `spec.action` is set to `STOP`.

On the service side, I added the handling needed to:

- call Kafka Connect stop for the targeted connector
- return an accepted response consistent with the other connector state operations
- keep the implementation aligned with the current service and client structure

## Tests

I added service-level coverage for the new stop flow in `ConnectorServiceTest`.

I also added an integration test in `ConnectorIntegrationTest` to verify the end-to-end `change-state` flow with `STOP`, including the transition from `RUNNING` to `STOPPED` on Kafka Connect.

## Remark

`STOP` is supported through the shared `change-state` API, which avoids adding a duplicated controller entrypoint for the same underlying behavior.

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/controller/connect/ConnectorController.java
+++ b/src/main/java/com/michelin/ns4kafka/controller/connect/ConnectorController.java
@@ -305,6 +305,7 @@ public class ConnectorController extends NamespacedResourceController {
             case RESTART -> response = connectorService.restart(ns, optionalConnector.get());
             case PAUSE -> response = connectorService.pause(ns, optionalConnector.get());
             case RESUME -> response = connectorService.resume(ns, optionalConnector.get());
+            case STOP -> response = connectorService.stop(ns, optionalConnector.get());
             default -> {
                 return Mono.error(new IllegalStateException(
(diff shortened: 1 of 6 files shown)


ThomasCAI-mlv (approved):
LGTM

PR #821, opened 2026-06-02
Add a functionality to fully reset connector offsets

AfrinKhan02:
# Context

Kafka Connect exposes dedicated offsets endpoints that allow administrators to inspect, alter, and reset connector offsets.

This PR adds support in Ns4Kafka for resetting connector offsets through the Kafka Connect REST API so Kafka admins can trigger the operation directly from the Ns4Kafka controller layer.

Referred API: https://docs.confluent.io/platform/current/connect/references/restapi.html#offsets

# Proposed Solution

Add a dedicated Ns4Kafka endpoint for connector offset reset:

`DELETE /api/namespaces/{namespace}/connectors/{connector}/offsets`

The endpoint delegates the reset operation to Kafka Connect and returns the exact success message coming back from Kafka Connect so the caller can see the result directly.

# Implementation

I added a dedicated controller endpoint to reset offsets for a connector owned by a namespace.

I then wired that endpoint through the connector service down to the Kafka Connect client using Kafka Connect's offsets reset endpoint:

`DELETE /connectors/{connector}/offsets`

On the Kafka Connect client side, I added a small response model to deserialize the message returned by Kafka Connect and forward it back through Ns4Kafka instead of returning an empty body.

The implementation keeps the change focused on the Full reset-offsets use case and aligned with the current controller and service patterns already used in Ns4Kafka.

# Tests

I added focused coverage for the new reset-offsets flow:

- controller tests for namespace ownership, connector not found, and successful reset through the dedicated endpoint
- service tests for delegation to Kafka Connect and propagation of the success response

The existing connector-focused tests were updated to validate the returned Kafka Connect message as part of the reset-offsets flow.

# Remark

This PR intentionally keeps reset-offsets as a dedicated controller endpoint and does not expose it through the generic connector `change-state` API.

Kafka Connect requires connectors to be stopped before offsets can be reset. Also this Pr is dedicated for Full Reset offset and not Partial /Alter offsets


ThomasCAI-mlv:
Hey @AfrinKhan02 can you rebase?

AfrinKhan02:
@ThomasCAI-mlv it's rebased, can you review now.

src/main/java/com/michelin/ns4kafka/controller/connect/ConnectorController.java
```diff
@@ -338,6 +339,30 @@ public Mono<MutableHttpResponse<ChangeConnectorState>> changeState(
...
+            return Mono.error(new ResourceValidationException(CONNECTOR, connector, invalidOwner(connector)));
+        }
+
+        Optional<Connector> optionalConnector = connectorService.findByName(ns, connector);
+
+        if (optionalConnector.isEmpty()) {
+            return Mono.just(HttpResponse.notFound());
+        }
+
+        return connectorService.resetOffsets(ns, optionalConnector.get());
```
loicgreffier: A check to verify that the connector is in the STOPPED state before running the reset might be needed. Currently, the user gets the following response:

```json
{
  "status": "Failed",
  "message": "Connectors must be in the STOPPED state before their offsets can be modified. This can be done for the specified connector by issuing a 'PUT' request to the '/connectors/abc.myDatagenConnector/stop' endpoint",
  "reason": "Internal Server Error",
  "details": {
    "causes": [
      "Connectors must be in the STOPPED state before their offsets can be modified. This can be done for the specified connector by issuing a 'PUT' request to the '/connectors/abc.myDatagenConnector/stop' endpoint"
    ]
  },
  "code": 500
}
```

The Ns4Kafka server returns a 500 error. We might want to override that response and instead suggest that the user stop the connector using Kafkactl before running the reset (we might need to add this status on Kafkactl side). This check can be done at controller level.

loicgreffier (commented):
@AfrinKhan02 Here is my review.

loicgreffier (approved):

PR #827, opened 2026-06-05
Change topic executor to deploy mode

ThomasCAI-mlv:


src/main/java/com/michelin/ns4kafka/controller/acl/AclController.java
```diff
@@ -151,11 +151,9 @@ public HttpResponse<AccessControlEntry> apply(
             throw new ResourceValidationException(accessControlEntry, validationErrors);
         }
 
-        // When the cluster manages RBAC, an ACL in "deleting" state is treated as non-existent,
-        // so that an apply-delete-apply flow transparently returns "created".
         Optional<AccessControlEntry> existingAcl = aclService
                 .findByName(namespace, accessControlEntry.getMetadata().getName())
-                .filter(existing -> !(aclService.isClusterManagingRbac(existing) && existing.isDeleting()));
+                .filter(existing -> !aclService.isClusterManagingRbac(existing));
```
loicgreffier: Any ACL on RBAC-managed cluster will return CREATED instead of UNCHANGED. `isClusterManagingRbac` condition needs to be removed, doesn't it?

src/test/java/com/michelin/ns4kafka/controller/connect/ConnectorControllerTest.java
```diff
@@ -794,65 +794,6 @@ void shouldChangeConnector() {
                 .verifyComplete();
     }
 
-    @Test
-    void shouldCreateConnectorWhenExistingIsDeleting() {
```
loicgreffier, line 798: Should the same test on the RBAC side be deleted as well?

loicgreffier (commented):
@ThomasCAI-mlv Here is my review

src/main/java/com/michelin/ns4kafka/controller/acl/AclController.java
ThomasCAI-mlv: Yes, you are right, I had removed the deleting check to be aligned with other executors, but these 2 conditions are linked, thus should be removed together

src/test/java/com/michelin/ns4kafka/controller/connect/ConnectorControllerTest.java
ThomasCAI-mlv: Due to the previous mistake, the test passed before. I remove it, thanks!

loicgreffier (approved):

PR #842, opened 2026-06-28
Add latest offsets and lag to consumer group listing

AfrinKhan02:
## Context

As a Kafka admin, when listing a consumer group I want to see the **current offset**, **end offset**, and **lag** per partition, so I can easily monitor how far each consumer group is behind on the topics it consumes.

Until now, the consumer group listing only exposed the committed `currentOffset` per partition. It did not return the partition end offset, nor the resulting lag. This enhancement is required on the Ns4Kafka side so that `kafkactl` can display this information when listing consumer groups (see [michelin/kafkactl#360](https://github.com/michelin/kafkactl/issues/360)).

## Proposed solution

Enrich the consumer group listing response with two additional fields per partition:

- `logEndOffset` - the latest (end) offset of the partition
- `lag` - the difference between the end offset and the committed current offset (`logEndOffset - currentOffset`)

The existing behavior is intentionally preserved: offsets are only returned for the owned-group listing when **exactly one** group matches the name filter.

## Implementation

- **Model** - Added `logEndOffset` and `lag` fields to `ConsumerGroup.ConsumerGroupOffset`.
- **Service** - In `ConsumerGroupService`:
  - Added a `getLogEndOffsets` helper that fetches the latest offsets for the committed partitions (returning an empty map when there are no partitions, and gracefully handling execution/interrupt exceptions, consistently with `getCommittedOffsets`).
  - `findByWildcardName` (owned groups) and `findExternalByWildcardName` (external groups) now retrieve the log end offsets for the committed partitions and pass them to `buildConsumerGroup`.
  - `buildConsumerGroup` populates `currentOffset`, `logEndOffset`, and the computed `lag` for each partition.

This reuses the existing `ConsumerGroupAsyncExecutor.getLogEndOffsets` admin call, so no new Kafka client logic was introduced.

## Tests

-  (`ConsumerGroupServiceTest`) - Mocked `getLogEndOffsets` and asserted `logEndOffset` and `lag` for both owned and external consumer group listing.
- (`ConsumerGroupIntegrationTest`) - Updated the expected `ConsumerGroupOffset` values to include the new end offset and lag fields when listing groups.

## Remark

The server-side rule of only returning offsets when a single group is queried is unchanged.

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/model/consumer/group/ConsumerGroup.java
+++ b/src/main/java/com/michelin/ns4kafka/model/consumer/group/ConsumerGroup.java
@@ -25,8 +25,6 @@ import io.micronaut.serde.annotation.Serdeable;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.NotNull;
-import java.util.ArrayList;
-import java.util.List;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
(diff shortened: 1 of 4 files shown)


loicgreffier (approved):
LGTM, merging

PR #858, opened 2026-07-14
Add Ns4Kafka validation before connector offsets reset

AfrinKhan02:
## Context

This PR adds the follow-up validation that was discussed during the review of https://github.com/michelin/ns4kafka/pull/821.

Today, when a user tries to reset offsets for a connector that is not stopped, Ns4Kafka delegates the call to Kafka Connect and surfaces the downstream failure. In practice, that ends up returning a Kafka Connect error instead of an Ns4Kafka-owned validation message.

## Proposed solution

Add an explicit validation in Ns4Kafka before sending the reset request to Kafka Connect.

If the connector is not in the `STOPPED` state, Ns4Kafka now fails earlier with its own `ResourceValidationException` and a dedicated message that tells the user to stop the connector first before retrying the reset.

## Implementation

I kept the change intentionally small and aligned with the existing code structure introduced by https://github.com/michelin/ns4kafka/pull/821.

The implementation adds a connector status check in the reset-offsets service path before calling Kafka Connect reset.

On the service side, the reset flow now:

- queries the current connector status from Kafka Connect
- validates that the connector is in the `STOPPED` state
- throws a `ResourceValidationException` with an Ns4Kafka-specific message when the connector is not stopped
- delegates to Kafka Connect reset only when the validation succeeds

## Tests

The test scope covers:

- successful connector offsets reset when the connector is already `STOPPED`
- validation failure when the connector is not stopped
- verification that Kafka Connect reset is not called when the precondition is not met

## Remark

This PR is specifically meant to cater the remaining review requirement from https://github.com/michelin/ns4kafka/pull/821.

### DIFF
--- a/src/main/java/com/michelin/ns4kafka/util/FormatErrorUtils.java
+++ b/src/main/java/com/michelin/ns4kafka/util/FormatErrorUtils.java
@@ -328,6 +328,23 @@ public class FormatErrorUtils {
     }
 
     /**
+     * Invalid reset offsets operation on connector.
+     *
+     * @param connector the connector
+     * @param targetState the target state
+     * @param currentState the current state
+     * @return the error message
+     */
+    public static String invalidConnectorResetOperation(String connector, String targetState, String currentState) {
+        return INVALID_OPERATION.formatted(
+                "reset offset",
+                String.format(
+                        "connector \"%s\" must be in the %s state before offsets can be reset. "
+                                + "Stop the connector first using ns4kafka API or Kafkactl, then retry the reset",
+                        connector, targetState));
+    }
+
+    /**
      * Invalid delete operation on consumer group.
      *
      * @param consumerGroup the consumer group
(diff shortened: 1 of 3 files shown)


src/main/java/com/michelin/ns4kafka/util/FormatErrorUtils.java
```diff
@@ -320,13 +321,29 @@ public static String invalidConsumerGroupOffsetNegative(String invalidOffsetValu
...
+     * @param connector the connector
+     * @param targetState the target state
+     * @return the error message
+     */
+    public static String invalidConnectorResetOperation(String connector, String targetState) {
+        return INVALID_OPERATION.formatted(
+                OPERATION_RESET_OFFSET,
+                String.format(
+                        "connector \"%s\" must be in the %s state before offsets can be reset. "
+                                + "Stop the connector first using ns4kafka API or Kafkactl, then retry the reset",
```
loicgreffier: We can suggest to simply stop the connector using `kafkactl connector stop` here.

loicgreffier (commented):
@AfrinKhan02 Here is my review. One minor comment.

loicgreffier (approved):

PR #884, opened 2026-09-15
Fix topic deletion when reapplied during synchronization

loicgreffier:


src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -296,7 +315,18 @@ public void createTopics(List<Topic> topics) {
...
      */
     public void alterTopics(List<Topic> targetTopics, Map<String, Topic> brokerTopics) {
-        Map<ConfigResource, Collection<AlterConfigOp>> topicConfigsToUpdate = targetTopics.stream()
+        Map<Boolean, List<Topic>> partitioned = targetTopics.stream()
+                .collect(Collectors.partitioningBy(
+                        topic -> brokerTopics.containsKey(topic.getMetadata().getName())));
+
+        // Topics missing from the broker cannot be altered, they are recreated instead
+        if (!partitioned.get(false).isEmpty()) {
+            createTopics(partitioned.get(false));
```
AfrinKhan02: Could we apply isUnchangedSinceLastApply before recreating missing topics too? This would avoid recreating a topic from an old update request when a newer deletion request already exists.

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -165,15 +165,34 @@ public List<String> listBrokerTopicNames() throws InterruptedException, Executio
...
+            }
+        }
 
         return managedClusterProperties
                 .getAdminClient()
-                .describeConfigs(topicNames.stream()
+                .describeConfigs(topicDescriptions.keySet().stream()
                         .map(topicName -> new ConfigResource(ConfigResource.Type.TOPIC, topicName))
                         .toList())
                 .all()
```
AfrinKhan02, line 198: Could we use the same per-topic handling for **describeConfigs** as you added for **describeTopics**? If a topic disappears between these calls, **.all().get()** still fails. Using .values() and handling **UnknownTopicOrPartitionException** individually would let the other updates continue.

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
```diff
@@ -296,7 +315,18 @@ public void createTopics(List<Topic> topics) {
...
     public void alterTopics(List<Topic> targetTopics, Map<String, Topic> brokerTopics) {
-        Map<ConfigResource, Collection<AlterConfigOp>> topicConfigsToUpdate = targetTopics.stream()
+        Map<Boolean, List<Topic>> partitioned = targetTopics.stream()
+                .collect(Collectors.partitioningBy(
+                        topic -> brokerTopics.containsKey(topic.getMetadata().getName())));
+
+        // Topics missing from the broker cannot be altered, they are recreated instead
+        if (!partitioned.get(false).isEmpty()) {
+            createTopics(partitioned.get(false));
+        }
```
AfrinKhan02, line 325: Could we apply **isUnchangedSinceLastApply** before recreating missing topics too? This would avoid recreating a topic from an old update request when a newer deletion request already exists.

Something like

```
List<Topic> missingTopics = partitioned.get(false).stream()
        .filter(this::isUnchangedSinceLastApply)
        .toList();

if (!missingTopics.isEmpty()) {
    createTopics(missingTopics);
}
```

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
loicgreffier: If a deletion lands between `describeTopics` and `describeConfigs`, then in the next synchronization loop, `describeTopics` will reject the update properly, and the topic will be created. IMHO, the fix on `describeTopics` only is acceptable for now, without adding too many "in case of" scenarios...

src/main/java/com/michelin/ns4kafka/service/executor/TopicAsyncExecutor.java
loicgreffier: `isUnchangedSinceLastApply` only checks whether the topic has been re-applied, so if a newer deletion request has landed, `isUnchangedSinceLastApply` won't prevent the creation. The topic will be deleted during the next synchronization

